PR #24, opened 2018-06-13
Added decision (aka rejection) reason data and API call.

katiemcgoff:


katiemcgoff:
@thorwolpert Can you please review and let me know if anything should change? This is the first API work I've done.

Also I think I need to write tests for this but I'm not sure and haven't investigated yet.

PR #37, opened 2018-06-19
Added two rows to decision_reason initial data via migration. DELETE DO NOT APPROVE

katiemcgoff:


katiemcgoff:
@thorwolpert Please review. I've added a data-only migration for two missing decision reasons per Debbie (descriptive, distinctive). The migration does simple inserts, no table creation or alters.

thorwolpert:
I think the up/down revisions need to be singular and in a chain. Can you change the down_revision to be the last in the list?

PR #55, opened 2018-06-25
Added fields for decision (reason, conflicts) to Name model.

katiemcgoff:
- added fields for decision to Name model
- added handling of "NRxxxxxxxx" format NR Number in validation function (rather than just "NR xxxxxxx")


katiemcgoff:
@thorwolpert Please review.

PR #59, opened 2018-06-26
added colin-api views #548

kialj876:


thorwolpert:
We can work on the views tomorrow. They'll need to be changed.

PR #62, opened 2018-06-27
Create view in Names to increase speed of conflicts API #548

kialj876:


thorwolpert:
Work on /name-examination/issue#548

thorwolpert:
ok, looks like all the issues are fixed.

PR #181, opened 2018-08-13
Changed automatic time entries from UTCNOW to simply NOW for PST/PDT.

katiemcgoff:
*Issue #, if available:* n/a

*Description of changes:*
Changed automatic time entries from UTCNOW to NOW - for PST/PDT instead of GMT.

This solves a few problems we have re. times being 8 hours ahead of actual local time, ie: we are storing UTC time (aka GMT, aka London time 8 hours ahead) in a timezone-less datetime field. 

Additionally this becomes a problem with the NRO Sync script, which looks for records recently updated (within the past 10 minutes). However any new updates in the database are 8 hours in the future, so it will take 8 hours (plus 10 minutes) before they can be picked up by the script. This change fixes that issue as well.

### DIFF
--- a/api/namex/models/comment.py
+++ b/api/namex/models/comment.py
@@ -9,7 +9,7 @@ class Comment(db.Model):
 
     id = db.Column(db.Integer, primary_key=True)
     comment = db.Column(db.String(4096))
-    timestamp = db.Column('timestamp', db.DateTime, default=datetime.utcnow)
+    timestamp = db.Column('timestamp', db.DateTime, default=datetime.now)
 
     # parent keys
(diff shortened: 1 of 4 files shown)


thorwolpert:
@katiemcgoff gov't generally uses UTC, as they have DCs in Kamloops (PST), Calgary (MDT), and other cloud sites. Generally it is made local during display to the user. UTC also doesn't do daylight savings, so there is no roll forward / back issues.

katiemcgoff:
@thorwolpert Okay I'll add a task to a) convert it when displaying in front-end, and b) use UTC time when checking in NRO sync for records to process.

PR #306, opened 2018-09-20
Pytests added for colin-api + r-words

kialj876:
*Issue #, if available:*564

*Description of changes:* 
- colin-api refactored into testable methods - all are tested in pytests
- restricted words pytests created to test for basic functionality


katiemcgoff:
Apparently RSA keys and private key values are already checked in elsewhere, so if this is a concern, it's a bridge we've already crossed so no NEW security problem via these commits.

PR #387, opened 2018-10-30
Add an index to the last_update and state_cd columns of the requests …

perkinss:
…table

*Issue #, if available: bcgov/name-examination#1081*

*Description of changes:*
Migration to create an index on the requests table to improve search times on the "completed" status codes and last_update (descending)

According to this sql, after running the queries, the ratios on that index improved significantly over several thousand updates and subsequent queries on the `requests` table.
https://gist.github.com/perkinss/686a4270195b2de5024bbc405d8c921f


katiemcgoff:
@perkinss Is this still in play? I read the comments in the ticket and I'm not clear whether this is still going forward. If it is, I approve.

perkinss:
@katiemcgoff I had some questions that I answered.  There are cases where they look earlier than the last 24 hours.  
So it's ok then.

PR #409, opened 2018-11-09
backend for filter state on completed on search page + pm test update

kialj876:
*Issue #, if available:NA*

*Description of changes:*
- updated pm tests for new data in prod (old test nr was wiped)
- updated pm tests for searching on 'completed' in state filter on search filter
- updated backend api for searching on 'completed' in state filter on search filter

### DIFF
--- a/api/namex/resources/requests.py
+++ b/api/namex/resources/requests.py
@@ -171,7 +171,6 @@ class Requests(Resource):
         # queue must be a list of states
         queue = request.args.get('queue', None)
         if queue:
-            states = queue.upper()
             if queue == 'COMPLETED':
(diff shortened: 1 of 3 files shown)


api/namex/resources/requests.py
```diff
@@ -171,6 +171,9 @@ def get(*args, **kwargs):
         # queue must be a list of states
         queue = request.args.get('queue', None)
         if queue:
+            states = queue.upper()
```
thorwolpert: -             states = queue.upper()

thorwolpert: Please remove ** states = queue.upper()** as it is not used

PR #419, opened 2018-11-15
1159 - we now handle more complex cases

ericminio:
*Issue #, if available:*

*Description of changes:*
use urllib in prod
explore writing tests the @pytest.mark.parametrize way
handle more exact-match cases

### DIFF
--- a/api/namex/resources/exact_match.py
+++ b/api/namex/resources/exact_match.py
@@ -1,9 +1,9 @@
 from flask import jsonify, request
 from flask_restplus import Resource, Namespace, cors
 from namex.utils.util import cors_preflight
-import requests
 import json
 from namex import jwt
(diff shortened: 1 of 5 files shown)


api/namex/resources/exact_match.py
```diff
@@ -24,10 +24,10 @@ def get():
               'sow=false' + \
               '&df=name_exact_match' + \
               '&wt=json' + \
-              '&q=' + query
-        results = requests.get(url)
-        text = results.text
-        answer = json.loads(text)
+              '&q=' + urllib.parse.quote(query)
+        print(url)
```
thorwolpert: consider using *current_app.logger.debug* (or info, warning, error) so that it logs out correctly on the pods.

PR #429, opened 2018-11-22
fix issue #18

ericminio:
office_vw now only includes not null delivery_addr_id and null end_event_id

### DIFF
--- a/nro-legacy/requirements.txt
+++ b/nro-legacy/requirements.txt
@@ -1,5 +1,3 @@
--r prod.txt
-
 pytest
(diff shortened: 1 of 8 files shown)


katiemcgoff:
Re. the requirements files, it would be better to have a single requirements.txt file since for this, it's only ever for dev (there is no "prod" scenario) and it's a bit out of the ordinary to not see a requirements.txt file and ONLY see a requirements folder.

After that change is made I will approve and merge. The SQL itself looks fine.

PR #432, opened 2018-11-28
20 - remove updates to nro request table

ericminio:
*Description of changes:*
we now preserve previous request id when updating request

### DIFF
--- a/api/namex/services/nro/change_nr.py
+++ b/api/namex/services/nro/change_nr.py
@@ -120,18 +120,6 @@ def _update_request(oracle_cursor, nr, event_id, change_flags):
                               admin_comment=row[20]
                               )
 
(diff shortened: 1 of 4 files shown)


thorwolpert:
@katiemcgoff Could you look at this, this removes code you added to support edit? I don't know which issue this is for, maybe @ericminio or @LJTrent could add that in. It may need to stay in to support edit, but may have been over-writing when not needed (just guessing as I know Lorna was checking into an issue where the previous_nr was missing and causing issues in NRO)

katiemcgoff:
@LJTrent Since we are removing this functionality, I don't think there's a reason to keep the front-end edit behaviour as well. Can you think of a reason to keep the edited field?

PR #447, opened 2018-12-03
24 - Condition view displays associated words

ericminio:


LJTrent:
The feedback from examiners  means there is  more work to be done before it can be shipped.

kialj876:
we will have 1 table instead of 2

PR #460, opened 2018-12-08
Add multi-word concatenation to the synonyms list

perkinss:
*Issue # bcgov/entity#52*

*Description of changes:*
To support finding synonyms that consist of more than one word, such as "Four two zero" matching "Cannabis". 
Solution @kialj876 and I collaborated on was to concatenate multi-word synonyms and add them to the list of single word synonyms.  So now, for a name like "Joes four two zero" instead of searching just for *four, two, zero* we search for *four, two, zero, fourtwo, fourtwozero, twozero*

- Wrote tests for expected output from synonym search
- Made the tests pass by updating the solr api method that finds synonym candidates from the name tokens
- Updated related tests that started to fail.

### DIFF
--- a/api/namex/analytics/solr.py
+++ b/api/namex/analytics/solr.py
@@ -302,7 +302,27 @@ class SolrQueries:
 
             previous_token = token
 
+        candidates.extend(cls._get_multiples(candidates))
+
         return candidates
+
+
+    @classmethod
+    def _get_multiples(cls, candidates):
+
+        if len(candidates) < 2:
+            return []
+
+        multiples = []
+
+        for x in range(len(candidates)):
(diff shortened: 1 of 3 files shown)


perkinss:
This looks like it's going to work ok.

api/namex/analytics/solr.py
```diff
@@ -302,13 +302,13 @@ def _parse_for_synonym_candidates(cls, tokens: List[str]) -> List[str]:
 
             previous_token = token
 
-        candidates.extend(cls._get_multiples(candidates))
+        candidates.extend(cls._get_concatenated_terms(candidates))
```
perkinss: good names are important. :)

PR #479, opened 2018-12-28
Issue#73 more history for tab

perkinss:
NOTE* Only this pr (479), or the alternative (480) should be accepted, but NOT BOTH!!

*Issue #, if available:* bcgov/entity#73

*Description of changes:*

- Update the solr schema for names core, for the "exact_match" field.  Which is actually a "same or similar" field.
-  Add tests to match up same and similar vs slightly different.
- Update the API to have a separate call to histories.


perkinss:
Awaiting feedback from business end to discover which PR should be merged.

PR #480, opened 2018-12-28
Version 2: Update solr names schema for exact match and do an exact-match histor…

perkinss:
**NOTE** Only this pr (480), or the alternative (479) should be accepted, but NOT BOTH!!

*Issue #, if available:* namex/entity#73

*Description of changes:*
- update solr names core with an exact_match field to match the 'exact_match' found in the possible.conflicts core.
- add test 
- update API to have a new histories endpoint.


perkinss:
Closing because this is not wanted (yet)

PR #483, opened 2019-01-03
Added a statement to the where clause wherever address is selected to…

LJTrent:
… ensure null addr_id is not allowed.

*Issue #, if available:*
https://github.com/bcgov/entity/issues/18

*Description of changes:*

Added addir_id IS NOT NULL to all address select statements in the colin api to eliminate error.


LJTrent:
I am closing this pull request as I need to finish the tests that test the xpro and registered office addresses do not return a null which is causing an error in the application.  I will complete these tests next week and create a new pull request.

PR #497, opened 2019-01-19
Correction to the view and updated tests

LJTrent:
*Issue #, if available:*

*Description of changes:*
Updated the solr conflict view for colin to include assumed names.
Added the missing end_event_id is null for xpros.


LJTrent:
Associated with https://github.com/bcgov/entity/issues/105

PR #498, opened 2019-01-20
Bug fix - upon reset, clear decision data in NRO but keep in Namex.

katiemcgoff:
*Issue #, if available:* bcgov/entity#110

*Description of changes:*
The decision data in NRO is cleared by calling name_examination proc to set to H. It used to call that until most recent changes to keep decision data rather than clear it. Now we call the proc but keep the data in Namex.

Side effect of this difference between name data in NRO vs. Namex is the blank data is synced back to Namex on next action, which is most likely examination/set to INPROGRESS, so mods were needed to NRO utils as well. The function was fairly rough so these changes improve it, with additional unit tests.


thorwolpert:
Is this a valid internal state change, as opposed to another flag that seems to be temporarily set in the tombstone record?

thorwolpert:
Tests for new code passes
Entire unit, non-integration test suite - green
Merging PR

PR #500, opened 2019-01-21
Handle Re-open bug - data being cleared from NRO upon Examine

katiemcgoff:
*Issue #, if available:* bcgov/entity#110

*Description of changes:*
Added hasBeenReset flag to API input/output (schema + json) and moved logic to after data load via schema, so that it is not overwritten if manually set (reset).

Corresponding front-end changes.


thorwolpert:
test suite cleared locally, letting roll to dev

PR #511, opened 2019-01-31
54 remove name choice

perkinss:
*Issue #, if available:*
bcgov/entity#54

*Description of changes:*
Check whether new name is blank before adding a new row to name_instance.


perkinss:
uhg, how did all those old changes get in there.  Starting over.

PR #512, opened 2019-01-31
Allow a name to be removed from nro name_instance table

perkinss:
*Issue #, if available:*
bcgov/entity#54

*Description of changes:*
Check whether new name is blank before adding a new row to the name_instance table in nro.

### DIFF
--- a/api/namex/models/name.py
+++ b/api/namex/models/name.py
@@ -66,7 +66,7 @@ class Name(db.Model):
     def delete_from_db(self):
         db.session.delete(self)
         db.session.commit()
-        
+
 
 class NameSchema(ma.ModelSchema):
     class Meta:
--- a/api/namex/models/name.py
+++ b/api/namex/models/name.py
@@ -67,6 +67,9 @@ class Name(db.Model):
         db.session.delete(self)
         db.session.commit()
 
+    def __init__(self, *args, **kwargs):
+        pass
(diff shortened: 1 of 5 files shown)


katiemcgoff:
I believe there is an api change needed too - current code just blanks out name but doesn't remove the name record. The record needs to be removed from Namex db, otherwise it's considered a name to be actioned on - ie: if name 3 is "removed" (just blanked out), user can reject 1, reject 2, and now we're on name 3 trying to make a decision on a blank name. This is a bug.

perkinss:
Oh! thanks for catching that
I'll look into it

katiemcgoff:
Sorry, I see now that you posted the options for the UI for this, but I didn't see that note (among the many other ticket update emails). I don't think this solution is intuitive - disabling but no messaging. I think your first option would be better - validation error on choice 2. In which case, disabling choice 3 shouldn't necessarily be disabled, that's also confusing. Simple validation message, like choice 1, would be best I think. 

If you don't agree, let's give this to Scott to go through UX stuff.

perkinss:
Working now

perkinss:
Oh doh, this is on the wrong repo

katiemcgoff:
There is one remaining issue, in postgres now, due to the changed behaviour. Closing this PR.

PR #522, opened 2019-02-14
104/147 - stems before synonym, number fix, bug on filtering conflict…

kialj876:
… list fix

*Issue #, if available:bcgov/entity#104 bcgov/entity#147*

*Description of changes:*
- synonyms are found based on their stems ('trucking' will match on 'truck' synonym list) 
- missed results are added to the bottom of the list instead of being filtered out
- less solr calls made from solr.py (speeds up conflict search)
- when doing a wildcard search proximity search does not run (increases speed and visually more clear)
- numbers are no longer split from letters in conflict search (5th = 5th/fifth/5, 5th != 5 th)
- pytests added

- 2 pytests fail locally because I don't have oracle running
- 1 pytest fails because solr-admin-app stems needs to rerun stems script after this pr is accepted


solr-synonyms-api/migrations/versions/d8e9c4ee2e74_.py
```diff
@@ -0,0 +1,77 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('synonym_bak')
+    op.drop_table('synonym_audit_bak')
+    op.drop_table('synonym_orig')
+    op.drop_table('synonym_audit')
+    op.alter_column('synonym', 'stems_text',
```
katiemcgoff: Unfortuntately this migration only alters the stems_text column by making it not nullable, whereas I think you wanted to have it create the stems_text column. It's done this because it was created (I'm assuming) at the dev solr database. So it saw that the column already existed, but noted the difference - nullable (the existing column) vs. your new code (not nullable) and made that change. Also, do you want the column to be not nullable? I'm guessing not - that means it can never be empty. OR you have to set it to default to empty string ("").

This migration will also drop the synonyms_audit table (line 24) which I'm guessing is not the intention. Unless it is in which case go for it.

We should also probably not use a migration to delete the _bak, _audit_bak, and _orig tables because they don't necessarily exist in test and production. If this migration fails to run because of anything, such as trying to drop a table that doesn't exist, it will not run at all and the column will not be created.

We're obviously in a bit of a messy situation right now because of all the table management happening manually up until this point, rather than migrations, but after this, we will be in a much better place for managing via the ORM (SQLAlchemy).

PR #532, opened 2019-02-22
Solr-API-app

rarmitag:
*Issue #, if available:* - BCGOV/entity#24

Completed app:

Synonyms-
  added stems_text column for searches

Restricted words - 
  4th table created to display and edit.

Decision Reason

Audit tables for each editable page

Two databases queried for data


solr-synonyms-api/app.py
```diff
@@ -10,4 +10,5 @@
 # Listen on all interfaces, and the catalog Python container expects the application to be on 8080.
 application = synonyms.create_application()
 if __name__ == '__main__':
-    application.run(host='0.0.0.0', port=8080, debug=True)
+    #application.run(host='0.0.0.0', port=8080, debug=True)
```
katiemcgoff: Should this be changed? I don't think so because I'm guessing it was/is only for local, but I don't know what effect it might have on running in the pods.

solr-synonyms-api/app.py
```diff
@@ -10,4 +10,5 @@
...
+    application.run(host='127.0.0.1', port=8080, debug=True)
```
katiemcgoff: Should this be changed? I don't think so because I'm guessing it was/is only for local, but I don't know what effect it might have on running in the pods.

thorwolpert: Yes, it should be set to 0.0.0.0

PR #546, opened 2019-03-08
Added consumption date and corp_num of entity that was created

stevenc987:
Signed-off-by: IDIR\schen <steven.chen@gov.bc.ca>

*Issue #, if available:*

*Description of changes:*


api/migrations/versions/28da946c5e79_.py
```diff
@@ -79,7 +79,8 @@ def upgrade():
     sa.Column('name', sa.String(length=1024), nullable=True),
     sa.Column('state', sa.String(length=15), nullable=True),
     sa.Column('choice', sa.Integer(), nullable=True),
-    sa.Column('consumption_date', sa.DateTime(), nullable=True),
+    sa.Column('consumption_date', sa.DateTime(), nullable=True),	
+	sa.Column('corp_num', sa.String(length=10), nullable=True),
     sa.Column('remote_name_id', sa.BigInteger(), nullable=True),
```
katiemcgoff: Hi Steven, this is a migration script, meaning it's called one time (upon deployment) to make changes to the database. This was already called in the beginning of the project. So it will not be re-run, and we do not add to them. So please remove this code change, and I can show you how to create a new migration script - it's a python thing.

nro-legacy/sql/object/names/namex/package/namex_pkb.sql
```diff
@@ -63,7 +63,7 @@ CREATE OR REPLACE PACKAGE BODY NAMEX.namex AS
 			IF (row_state_type_cd IN ('C', 'D')
 			    AND 
 			    row_transaction_type_cd IN 
-			         ('ADMIN', 'NRREQ', 'RESUBMIT', 'CANCL', 'MODIF', 'CORRT', 'UPDPR')
+			         ('ADMIN', 'NRREQ', 'RESUBMIT', 'CANCL', 'MODIF', 'CORRT', 'UPDPR', 'CONSUME', 'EXPIR')
```
katiemcgoff: As discussed in slack, please remove EXPIR - turns out we don't need this.

api/namex/models/name.py
```diff
@@ -71,8 +73,8 @@ def delete_from_db(self):
 class NameSchema(ma.ModelSchema):
     class Meta:
         model = Name
-        fields = ('name', 'state', 'choice', 'designation', 'consumptionDate', 'conflict1', 'conflict2', 'conflict3',
-                  'conflict1_num', 'conflict2_num', 'conflict3_num', 'decision_text')
+        fields = ('name', 'state', 'choice', 'designation', 'consumptionDate', 'corpNum', 'conflict1', 'conflict2',
+                  'conflict3', 'conflict1_num', 'conflict2_num', 'conflict3_num', 'decision_text')
```
katiemcgoff: Having real consumptionDate data going to/from the front-end now causes validation errors when unrelated data is sent back to the API from the front-end. To prevent this, we need to remove consumptionDate from the json data (as_dict()). It's not used anywhere in the front-end yet so this is probably the best way to prevent bugs until the front-end can be managed.

Similarly, we also need to remove corpNum from as_dict() and NamesSchema. It will get overwritten during saves if it remains inthe NamesSchema. 

That means a few tests will also have to be changed to remove that data from the schema, until we can put it back.

jobs/nro-update/tests/test_nro_datapump.py
```diff
@@ -1,17 +1,15 @@
...
-import pytz
 
 from nro.nro_datapump import nro_data_pump_update, create_expiry_date
 from namex.models import Request, Name, State, User
 
 
 expiry_date_test_data = [
     ('using epoch utc',            # test descriptive name
-     datetime.utcfromtimestamp(0), # start date - time
+     datetime(1970, 1, 1, 00, 00, tzinfo=timezone('US/Pacific', )), # start date - time
```
katiemcgoff: What is the purpose of changing these dates from datetime.utcfromtimestamp(0) to the 1970 date? You've done it in a few places in this test file.

PR #547, opened 2019-03-11
missed results bug fix

kialj876:

*Issue #, if available:bcgov/entity#222*

*Description of changes:*
- fixed missed results bug
- added pytest to secure it


api/tests/python/end_points/test_synonym_match.py
```diff
@@ -76,12 +76,11 @@ def verify(data, expected=None, not_expected=None):
 
     verified = False
     print(data['names'])
-
+    print('EXPECTED ', expected)
+    print('NOT EXPECTED ', not_expected)
```
katiemcgoff: should remove print statements

PR #567, opened 2019-04-29
Added CONSUME transaction handling in namex feeder

katiemcgoff:
*Issue #, if available:* bcgov/entity#508

*Description of changes:*
Added CONSUME transaction handling in namex feeder; handle in a way that does not include unexpected COMPLETED transactions.

Note: CONSUME transaction type was previously added to condition, but is never executed - see ticket for notes.


WalterMoar:
Script runs, change deployed to NAMESD.

PR #606, opened 2020-01-08
2114 word class model

LJTrent:
*Issue #, if available:*
Associated with 2114.  Just the model though. 

*Description of changes:*


api/namex/models/word_classification.py
```diff
@@ -0,0 +1,49 @@
...
+    word = db.Column('word', db.String(1024), nullable=False, index=True)
+    lastNameUsed = db.Column('last_name_used',db.String(1024))
+    lastPrepName = db.Column('last_prep_name',db.String(1024))
+    frequency = db.Column('frequency', db.BIGINT)
+    approvedBy = db.Column('approved_by', db.Integer, db.ForeignKey('users.id'))
+    approvedDate = db.Column('approved_dt', db.DateTime(timezone=True), default=datetime.utcnow)
+    startDate = db.Column('start_dt', db.DateTime(timezone=True), default=datetime.utcnow)
+    endDate = db.Column('end_dt', db.DateTime(timezone=True), default=datetime.utcnow)
+    lastUpdatedBy = db.Column('last_updated_by', db.Integer, db.ForeignKey('users.id'))
+    lastUpdateDate =  db.Column('last_update_dt', db.DateTime(timezone=True), default=datetime.utcnow)
```
katiemcgoff: You might want to add an 'onupdate' value so that this updates automatically every time the record is changed (if that's the functionality you intended. Here's the namex request version:
```
    lastUpdate = db.Column('last_update', db.DateTime(timezone=True), default=datetime.utcnow, onupdate=datetime.utcnow)
```

api/namex/models/word_classification.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def save_to_db(self):
+        db.session.add(self)
+        db.session.commit()
+
+    def save_to_session(self):
+        db.session.add(self)
+
+    def delete_from_db(self):
+        raise BusinessException()
```
katiemcgoff: So this won't actually delete the record, only raise an error. Are you sure you want the method delete_from_db() at all then, if it doesn't actually do anything?

api/namex/models/word_classification.py
LJTrent: I will do that.  Thanks.

api/namex/models/word_classification.py
LJTrent: I know it wont delete. I had it there because I think there are circumstances when we do want to delete it but you are right I should take it out and inly add it back when I figure it out.

So yes I will remove it.  Thx.

api/namex/models/word_classification.py
```diff
@@ -0,0 +1,49 @@
...
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    classification = db.Column('word_classification',db.String(4),default='NONE',nullable=False,index=True)
+    word = db.Column('word', db.String(1024), nullable=False, index=True)
+    lastNameUsed = db.Column('last_name_used',db.String(1024))
+    lastPrepName = db.Column('last_prep_name',db.String(1024))
+    frequency = db.Column('frequency', db.BIGINT)
+    approvedBy = db.Column('approved_by', db.Integer, db.ForeignKey('users.id'))
+    approvedDate = db.Column('approved_dt', db.DateTime(timezone=True), default=datetime.utcnow)
+    startDate = db.Column('start_dt', db.DateTime(timezone=True), default=datetime.utcnow)
+    endDate = db.Column('end_dt', db.DateTime(timezone=True), default=datetime.utcnow)
```
katiemcgoff: When you first create a record, this will populate the endDate value with the current date. Do you want that? If you leave this in, and you want endDate to be blank upon insert, you will have to specifically say endDate=None. Same case with approvedDate above but I'm not sure I know the functionality around that one.

PR #666, opened 2020-04-30
3532 sort designations by language

areyeslo:
*#3532, Sort designations by Language:*

*Description of changes:*
-Sort by Language, then by length
-Fixes related to designation variables in GetSynonymListsMixin that have to be in GetDesignationsListsMixin. In GetSynonymListsMixin we leave just prefixes, substitution, stop words
-Removed unused variables in _clean_name_words, we are sending all_designations instead of splitting in any/end (this was commented, just removing comments)
-Returning list instead of list of tuples in get_designation_any_in_name


bcdevlucas:
Looks fine to me.

PR #668, opened 2020-05-01
Removing variable all_designation_user_no_periods storing designation…

areyeslo:
#3476, Handle Missing Period on Designation Mismatch and Move Designation:*

*Description of changes:*
-Removing variable (_all_designations_user_no_periods) storing designations without periods. Now LTD OR LLC are not longer considered designations because they have no period included. Just LTD. , L.L.C. are valid designations.


LJTrent:
replaced with my branch to resolve merge cnflict

PR #707, opened 2020-05-26
3688 v2 fix name cleaning bugs discovered pytests

areyeslo:
*#3688, Fix name cleaning bugs discovered when running pytests:*

*Description of changes:*

- Change to upper case responses shown in FE
- Change special and words to avoid to show in FE as compound when they are more than one word such as "BIG BROS" instead of showing "BIG","BROS"
- Change response from None to ProcedureResult type for well_formed, unclassified words and check limit.
- Removed unused misplaced_any_designation and misplaced_all_designation
- Refactoring in clean name to avoid calling regex-prefix and remove_french twice for each request.
- Fixed remove_french to split the name as long as there is a designation before the slash or hyphen
- Remove numbers inside parenthesis with anything else inside such as (YR 2008) or (SOMETHING 2008).
- Exclude designations to be transformed after setting together prefixes and following word. For instance, CO-OP was transformed to COOP, now it keeps the original form.
- Search conflict stops when it finds a exact match.
- Stand-alone words are retrieved from database instead of hard coded. Database needs to be updated to include enterprises in stand-alone words:
  **update synonym
  set synonyms_text='bc, holdings, ventures, solutions, enterprise, enterprises, industries'
  where category='Stand-Alone'

  update synonym
  set stems_text='bc, holdings, ventures, solutions, enterprise, enterprises, industries' 
  where category='Stand-Alone'**

- All test cases were successful except test_set_name_regex_separated_ordinals (We need to get an insight for possible changes in here).

### DIFF
--- a/api/namex/resources/auto_analyse/analysis_issues.py
+++ b/api/namex/resources/auto_analyse/analysis_issues.py
@@ -40,8 +40,8 @@ class AnalysisResponseIssue:
             return []  # This method should always return a list
 
         try:
-            converted_list = list(map(lambda d: d.lower() if isinstance(d, str) else '', str_list)) \
-                if convert else list(map(lambda d: d.lower(), str_list))
+            converted_list = list(map(lambda d: d.upper() if isinstance(d, str) else '', str_list)) \
+                if convert else list(map(lambda d: d.upper(), str_list))
         except Exception as err:
             print('List is not a list of strings ' + repr(err))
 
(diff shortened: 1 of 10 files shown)


areyeslo:
Currently in master, the function check_word_special_use is using name_tokens and original_name, the name tokens is used to iterate over the list of words for the processed name and find any special word.  In this change, we provide the string of processed name instead of list and search for compound words. In brief, before we were using a list of words and now string, but both keep the process name.

In regards to remove-french, we encountered several names that were not handled by the current function, so we realized that we needed to include the rule for having a designation before the slash. With this change all the test cases that were failing now passed.

PR #710, opened 2020-05-28
3767 v2 handle more than one correct designation

areyeslo:
*#3767, Handle more than once correct end designation:*

*Description of changes:*

- Creation of new response that substitute to designation_misplaced when there are more than one end designation. The new response send just the end designations included in the name and send to the user the following response if provided "LTD. ARMSTRONG CORPORATION PLUMBING INC.":
'There can be only one designation. You must choose either <b>LTD.</b>  or  <b>CORPORATION</b>  or  <b>INC.</b>'

- Test cases are updated.


LJTrent:
Doesn't handle designations in the middle ABC LTD. PLUMBING INC.  gives a different response. Lets see how this plays out with the front-end. Does handle it when they are both at the end.

PR #755, opened 2020-07-17
feature/4261-Existing-NR-Search

bcdevlucas:
*Issue #, if available:*

*Description of changes:*


bcdevlucas:
> Ensure that if there are no results that the front-end can turn this into a user message not just a 400 or 404.

There is a message attached. The frontend needs to handle the error.

PR #777, opened 2020-08-12
feature/4266-4267-4269-4484-4570-Existing-NR-Patch

bcdevlucas:
*Issue #, if available:*

*Description of changes:*


api/namex/resources/name_requests/name_request.py
```diff
@@ -250,72 +259,152 @@ def handle_patch_edit(self, nr_model):
...
+        if nr_name and nr_applicant:
+            payment_request = {
+                'paymentInfo': {
+                    'methodOfPayment': 'CC',
+                },
+                'filingInfo': {
+                    'date': date.today(),
+                    'filingTypes': [
+                        {
+                            'filingDescription': 'NM620: ' + nr_name.name,
```
LJTrent, line 291: Please set the filing description to null so the receipt does not have the fee code NM620 and also so the receipt does not have the name as a header.  This in on advice from Sumesh.

api/namex/resources/name_requests/name_request.py
```diff
@@ -250,72 +259,152 @@ def handle_patch_edit(self, nr_model):
...
+            if not payment:
+                raise PaymentServiceException(MSG_ERROR_CREATING_RESOURCE)
+
+            # Update the name request with the payment id
+            # nr_draft.paymentToken = str(payment.id)
+            nr_model.payment_token = str(payment.id)
+            # nr_model.payment_completion_date = ''
 
+            nr_model.priorityCd = 'Y'
+            # nr_model.priorityDate = ''
```
LJTrent, line 338: Need to set the priorityDate now()  utc

api/namex/resources/name_requests/name_request.py
```diff
@@ -250,72 +259,152 @@ def handle_patch_edit(self, nr_model):
...
-
         if nr_model.submitCount < 3:
-            # Update submit count
-            nr_model = nr_svc.update_request_submit_count(nr_model)
-            # Extend expiry date by (default) 56 days
-            nr_model = nr_svc.extend_expiry_date(nr_model)
-            # TODO: If request action is REH or REST extend by 1 year (+ 56 default) days
-            # TODO: Where does the request action come from in this scenario?
-            # nr_model = nr_svc.extend_expiry_date(nr_model, (datetime.utcnow() + relativedelta(years=1)))
+            if nr_svc.request_action in [RequestAction.REH.value, RequestAction.REST.value]:
```
LJTrent, line 362: sorry the request_actions are REH(Restore from a Historical Name) and REN(Restore with a New Name). Please update.  I double-checked the front-end code.

api/namex/resources/name_requests/name_request.py
```diff
@@ -250,72 +259,152 @@ def handle_patch_edit(self, nr_model):
...
             self.update_records_in_network_services(nr_model)
+        else:
+            # TODO: Make a custom exception for this?
+            raise NameRequestException(message='Submit count maximum of 3 retries has been reached!')
 
-            # TODO: Raise an error if submitCount is greater than 3
         return nr_model
 
     def handle_patch_resend(self, nr_model):
         # This handles updates if the NR state is 'patchable'
```
LJTrent, line 383: I need to figure out how the resend is triggered in NRO so we know what field to update.  I will add this as another ticket. SO we will leave this for now in this PR.

api/namex/resources/name_requests/name_request.py
```diff
@@ -250,72 +259,152 @@ def handle_patch_edit(self, nr_model):
...
         nr_model = self.update_nr_fields(nr_model, nr_model.stateCd)
```
LJTrent, line 384: We will leave this for now as I need to figure out how a resend is triggered in NRO.

api/namex/resources/name_requests/resource.py
```diff
@@ -294,12 +285,15 @@ def update_request_in_nro(self, name_request, on_success=None):
                 })
 
             return self.on_nro_update_complete(name_request, on_success, nro_warnings)
-        # Handle any changes where ONLY state is changed
-        elif name_request.stateCd in [State.CANCELLED]:
-            # TODO: It might be a good idea to set an env var for this... as NRO can't run in local tests
-            nro_warnings = self.nro_service.change_nr(name_request, {
-                NROChangeFlags.STATE.value: True
-            })
+        elif name_request.stateCd in [State.COND_RESERVE, State.RESERVED, State.CONDITIONAL, State.APPROVED, State.CANCELLED]:
```
LJTrent, line 288: We dont save to NRO if the state is COND_RESERVE or RESERVED.  Please remove.

api/namex/resources/name_requests/resource.py
```diff
@@ -294,12 +285,15 @@ def update_request_in_nro(self, name_request, on_success=None):
...
+            if current_app.config.get('DISABLE_NAMEREQUEST_NRO_UPDATES', 0) == 1:
+                # Ignore update to NRO if NRO updates [DISABLE_NAMEREQUEST_NRO_UPDATES] are explicitly disabled in your .env
+                nro_warnings = None
+            else:
+                nro_warnings = self.nro_service.change_nr(name_request, {
+                    NROChangeFlags.REQUEST.value: True,
+                    NROChangeFlags.STATE.value: True
```
LJTrent, line 295: We only update the state in NRO when the state=CANCELLED. Not for APPROVED or CONDITIONAL. 
elif name_request.stateCd in [State.CONDITIONAL, State.APPROVED, State.CANCELLED]:

....
if State.State.CANCELLED: NROChangeFlag,STATE.value = True 
else: NROChangeFlags.REQUEST.value = True

nro_warnings = self.elf.nro_service.change_nr(name_request, {
                    NROChangeFlags}

api/namex/services/name_request/abstract_name_request.py
```diff
@@ -99,12 +113,15 @@ def next_state_code(self, val):
 
     @classmethod
     def set_request_type(cls, entity_type, request_action):
```
LJTrent, line 115: generally this is not supplied by the front-end and will set by the event listener on the request model except in the case of an XPRO/Assumed name.

LJTrent: Really not sure why you have this mapping.  I think it should be removed.  The request_type if it is supplied should just be included as any other field in the request attributes update (header) other wise the event listener is setting it.  please take this out.

api/namex/services/name_request/auto_analyse/__init__.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, **kwargs):
         self.values = kwargs.get('values', None)
 
 
+# TODO: Why can't we use the RequestAction enum in namex.constants
 class AnalysisRequestActions(str, Enum):
     NEW = 'NEW'  # Start a new business (NAME PROTECTION)
```
LJTrent, line 35: NEW does not mean Name PROTECTION

api/namex/services/name_request/name_request.py
```diff
@@ -217,8 +215,9 @@ def map_request_data(self, name_request, map_draft_attrs=False):
         # If this is a draft, set name request header attributes
         if map_draft_attrs:
             name_request = self.map_draft_attrs(name_request)
-            name_request = self.map_request_header_attrs(name_request)
-            name_request = self.map_request_comments(name_request)
+
+        name_request = self.map_request_header_attrs(name_request)
+        name_request = self.map_request_comments(name_request)
 
         try:
```
LJTrent, line 222: This try block doesnt do anything?

api/namex/services/name_request/name_request.py
```diff
@@ -250,10 +247,12 @@ def map_draft_attrs(self, name_request):
...
             name_request.userId = user_id
-            name_request.entity_type_cd = request_entity
-            name_request.request_action_cd = request_action
             name_request.submittedDate = datetime.utcnow()
-            name_request.requestTypeCd = self.set_request_type(request_entity, request_action)
+            name_request.entity_type_cd = request_entity
+
+            if request_action:
+                name_request.request_action_cd = request_action
+                name_request.requestTypeCd = self.set_request_type(name_request.entity_type_cd, request_action)
```
LJTrent, line 255: take this out. the vent listener does this already. Only if the front-end supplies the request type should you be settting it here like any other field supplied by the front-end in the payload:
 name_request.requestTypeCd = request_type_cd

api/namex/services/name_request/name_request.py
bcdevlucas: It does:
```
        try:
            if new_state_code == State.COND_RESERVE:
                name_request.consentFlag = 'Y'

            if new_state_code in [State.RESERVED, State.COND_RESERVE]:
                name_request.expirationDate = self.create_expiry_date(
                    start=name_request.submittedDate,
                    expires_in_days=56,
                    tz=timezone('UTC')
                )
        except Exception as err:
            raise MapRequestDataError(err)
```

api/namex/services/name_request/name_request.py
bcdevlucas: I will have to confirm that behaviour first. If it's working, without it I should be able to leave it out.

api/namex/resources/name_requests/name_request.py
bcdevlucas: Noted.

api/namex/services/name_request/abstract_name_request.py
bcdevlucas: I don't think we're on the same page. These properties are required when mapping data, and are not exclusive to the PATCH. They are also used for mapping PUT and POST data. Doesn't work without those changes in there, as the ENTITY uses request_action_cd but the frontend uses request_action. However the frontend does something weird which is returning the entity with request_action as the property instead of using request_action_cd. Like if the pay load looks like: 
```
{ request_action: 'blah', data: { <entity fields> }
but instead i get:
{ data: { <entity fields>, request_action: 'blah', request_action_cd: 'blah' } where request_action_cd is the actual value of the entity. Don't pass weird fields in the entity from the frontend. 

Anyway, without a change on the frontend, I need to check for the existence of either code. Not having this change breaks stuff.
```

LJTrent: We will remove the event listener for def update_request_action_entity_type in the request model. and use this set_request_type(cls, entity_type, request_action):  However, if the payload supplies the requestTypeCd we just set it and dont use the mapper.

api/namex/services/name_request/name_request.py
LJTrent: ok.  use the create_expiry_date to set the priority date.

api/namex/services/name_request/auto_analyse/__init__.py
LJTrent: later

PR #819, opened 2020-09-19
4844/fixes uat rejection testing stand alones

areyeslo:
*I4844 #, Fixes UAT Rejection Testing Stand Alones:*

*Description of changes:*
Search for exact stand alone word using synonyms_text instead of stems_Text.
Check if stand alone is having an additional descriptive or distinctive.
Allow numbers at the beginning of a name

### DIFF
--- a/api/namex/services/name_request/builders/name_analysis_builder.py
+++ b/api/namex/services/name_request/builders/name_analysis_builder.py
@@ -634,9 +634,9 @@ class NameAnalysisBuilder(AbstractNameAnalysisBuilder):
     def stand_alone_name_different(self, lst_dist_name1, lst_dist_name2, lst_desc_name1, lst_desc_name2,
                                    stand_alone_words):
         if self.is_standalone_name(lst_desc_name1, stand_alone_words) and self.is_standalone_name(lst_desc_name2,
-                                                                                                  stand_alone_words):
-            if lst_dist_name1.__len__() != lst_dist_name2.__len__() or lst_desc_name1.__len__() != lst_desc_name2.__len__():
-                return True
+                                                                                                  stand_alone_words) and (
+                lst_dist_name1.__len__() != lst_dist_name2.__len__() or lst_desc_name1.__len__() != lst_desc_name2.__len__()):
+            return True
 
         return False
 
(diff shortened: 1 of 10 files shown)


api/namex/services/name_request/builders/name_analysis_builder.py
```diff
@@ -635,7 +635,7 @@ def stand_alone_name_different(self, lst_dist_name1, lst_dist_name2, lst_desc_na
                                    stand_alone_words):
         if self.is_standalone_name(lst_desc_name1, stand_alone_words) and self.is_standalone_name(lst_desc_name2,
                                                                                                   stand_alone_words):
-            if lst_dist_name1.__len__() > lst_dist_name2.__len__() or lst_desc_name1.__len__() > lst_desc_name2.__len__():
+            if lst_dist_name1.__len__() != lst_dist_name2.__len__() or lst_desc_name1.__len__() != lst_desc_name2.__len__():
```
LJTrent: This feels like it should be in its own function and not embedded in this function since this check.rule an applied to all names.

api/namex/services/name_request/builders/name_analysis_builder.py
```diff
@@ -444,10 +444,12 @@ def get_most_similar_names(self, dict_highest_counter, matches, dist_substitutio
         list_desc = list(desc_synonym_dict.keys())
         if matches:
             syn_svc = self.synonym_service
+            nproc_svc = self.name_processing_service
             service = ProtectedNameAnalysisService()
             np_svc = service.name_processing_service
             wc_svc = service.word_classification_service
             token_svc = service.token_classifier_service
+            stand_alone_words = nproc_svc.get_stand_alone_words()
```
LJTrent, line 452: There should be a separate function to check for dist desc dist pattern as well as dist dec desc pattern where they are in the same category for all names not embeded in the stand-alones functions..

api/namex/services/name_request/builders/name_analysis_builder.py
```diff
@@ -634,9 +634,9 @@ def is_standalone_name(self, list_name, stand_alone_words):
     def stand_alone_name_different(self, lst_dist_name1, lst_dist_name2, lst_desc_name1, lst_desc_name2,
```
LJTrent, line 634: what does thsi method really do?  The name is unclear.

PR #821, opened 2020-09-23
feature/5008-Finalize-PATCH-Action-Logic

bcdevlucas:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/api/namex/models/request.py
+++ b/api/namex/models/request.py
@@ -112,6 +112,19 @@ class Request(db.Model):
 
     # properties
     @property
+    def has_consumed_name(self):
+        names = self.names.all()
+
+        def is_not_consumed(name):
+            no_corp_num = name.corpNum is None
+            no_consumption_date = name.consumptionDate is None
+
+            return no_corp_num and no_consumption_date
+
+        results = [is_not_consumed(n) is False for n in names]
+        return True in results
+
+    @property
     def latest_payment(self):
         payments = self.payments.all()
         return payments[0]
@@ -120,6 +133,14 @@ class Request(db.Model):
     def latest_payment_completion_date(self):
         payments = self.payments.all()
         return payments[0]
+
+    @property
+    def is_expired(self):
+        todays_date = datetime.utcnow().date()
+        expiry_date = self.expirationDate.date() if self.expirationDate else None
+        if expiry_date:
+            return todays_date > expiry_date
+        return False
 
     @property
     def source(self):
--- a/api/namex/resources/name_requests/api_models.py
+++ b/api/namex/resources/name_requests/api_models.py
@@ -33,7 +33,8 @@ name_model = api.model('name_model', {
     'designation': fields.String('Name designation based on entity type'),
     'conflict1_num': fields.String('The corp_num of the matching name'),
     'conflict1': fields.String('The matching corp name'),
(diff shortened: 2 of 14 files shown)


api/namex/models/request.py
```diff
@@ -111,6 +111,19 @@ class Request(db.Model):
     REQUEST_FURNISHED = 'Y'
 
     # properties
+    @property
+    def has_consumed_name(self):
```
LJTrent, line 115: hmm. wondering why this is not part of the names model instead?

api/namex/models/request.py
```diff
@@ -111,6 +111,19 @@ class Request(db.Model):
...
+        names = self.names.all()
+
+        def is_not_consumed(name):
```
LJTrent, line 118: same thing, should this be in the names model?

api/namex/resources/name_requests/api_models.py
```diff
@@ -33,7 +33,8 @@
     'designation': fields.String('Name designation based on entity type'),
     'conflict1_num': fields.String('The corp_num of the matching name'),
     'conflict1': fields.String('The matching corp name'),
-    'consent_words': fields.Nested(consent_model)
+    'consent_words': fields.Nested(consent_model),
+    'corpNum': fields.String('Corp Num')
```
LJTrent, line 37: what about  consumptionDate

api/namex/resources/name_requests/name_request.py
```diff
@@ -267,20 +267,37 @@ def patch(self, nr_id, nr_action):
             nr_svc.nr_num = nr_model.nrNum
             nr_svc.nr_id = nr_model.id
 
-            valid_states = State.VALID_STATES
+            # Only allow editing if the request is in certain valid states
+            request_editable_states = [
+                State.DRAFT,
+                State.RESERVED,
```
LJTrent, line 273: Why is reserved and cond-reserve editable.  We dont want to edit those from the patch?

api/namex/models/request.py
bcdevlucas: Convenience eg.
`
def display_cancel_action(nr_model=None):
    try:
        if (nr_model and nr_model.stateCd == State.CANCELLED) or (nr_model.stateCd not in State.CANCELLABLE_STATES):
            return False

        if nr_model and (nr_model.is_expired or nr_model.has_consumed_name):
            return False

        return True
    except Exception as err:
        raise NameRequestActionError(err)
`

api/namex/models/request.py
bcdevlucas: Same as my comment above.

api/namex/resources/name_requests/api_models.py
bcdevlucas: Not sure about consumptionDate... I didn't see that in the ticket.

api/namex/resources/name_requests/name_request.py
bcdevlucas: Maybe editable isn't the best word... I'm just checking to see what states should actually be 'handled' within the PATCH method.

api/namex/resources/name_requests/api_models.py
LJTrent: consumptionDate and corpNum define this as a consumed name.

PR #849, opened 2020-10-03
feature/5060 search exact match

areyeslo:
*5060  #, Search Exact Match:*

*Description of changes:*
Implementation of search exact match

### DIFF
--- a/solr-synonyms-api/synonyms/constants/__init__.py
+++ b/solr-synonyms-api/synonyms/constants/__init__.py
@@ -57,7 +57,7 @@ class EntityTypes(AbstractEnum):
     COOPERATIVE = 'CP'
     BENEFIT_COMPANY = 'BC'
     COMMUNITY_CONTRIBUTION_COMPANY = 'CC'
-    SOCIETY = 'SO'
+    SOCIETY = 'SOC'
     PRIVATE_ACT = 'PA'
     FINANCIAL_INSTITUTION = 'FI'
     PARISH = 'PAR'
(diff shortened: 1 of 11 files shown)


solr-synonyms-api/synonyms/constants/__init__.py
```diff
@@ -57,7 +57,7 @@ class EntityTypes(AbstractEnum):
     COOPERATIVE = 'CP'
     BENEFIT_COMPANY = 'BC'
     COMMUNITY_CONTRIBUTION_COMPANY = 'CC'
-    SOCIETY = 'SO'
+    SOCIETY = 'SOC'
```
LJTrent, line 60: SHould this list of entity types match the one in namx-api?  If so, can you not import from namex-api rather than duplicate it in synonyms?

solr-synonyms-api/synonyms/constants/__init__.py
```diff
@@ -98,6 +98,7 @@ class BCProtectedNameEntityTypes(AbstractEnum):
     LIMITED_LIABILITY_PARTNERSHIP = EntityTypes.LIMITED_LIABILITY_PARTNERSHIP.value
     COOPERATIVE = EntityTypes.COOPERATIVE.value
     COMMUNITY_CONTRIBUTION_COMPANY = EntityTypes.COMMUNITY_CONTRIBUTION_COMPANY.value
+    SOCIETY = EntityTypes.SOCIETY.value
```
LJTrent, line 101: same here. Why duplicate them.  Just import them where you need them from namex-api if they are the same.  Probably the constants in syonyms shoudl be reviewed and all duplication removed.  I will create a ticket for that.

PR #898, opened 2020-10-24
5279/Handle names with same category and longer than two words in length.

areyeslo:
*5279 #, Handle Name Where All Words Are Classified as All Dist or All Desc:*

*Description of changes:*
- Modify get_classification to replace for descriptive the last word if all words are distinctive. In the same way, replace for distinctive the first word if all words are descriptive.


kialj876 (approved):
Looks fine to me, but might want to consider adding a test or 2 to secure this new functionality. Also is there a reason for using ```.__len__()``` instead of the ```len()``` function?

areyeslo:
len(s) is a built-in Python method which returns the length of an object.  __len__() is a internal method  in objects which is called by len(s) method to return the length of an object.  So, when we call len(s) method, s.__len__() is executed to calculate the length.

kialj876:
thanks for adding in those pytests

PR #926, opened 2020-11-13
5593 change query search distinctives

areyeslo:
*bcgov/entity#5593*

*Description of changes:*
- Retrieve conflicts that include distinctive as second token followed by descriptive. Before just the first token was considered a valid distinctive.
- When getting synonyms do not include stand-alone words to avoid mixing them with regular synonyms.
- Name analysis was changed to be over the full name instead of dropping synonyms in the same category.
- Removing duplicated letters is done after getting synonyms.
- Added test cases for distinctive after first place

Test 1:
image

Test 2:
image

Test 3:
image

Test 4:
image

Test 5:
image

Test 6:
image

Regression test:
image

### DIFF
--- a/api/namex/models/request.py
+++ b/api/namex/models/request.py
@@ -498,10 +498,10 @@ class Request(db.Model):
             substitutions = '|'.join(map(str, special_characters_dist))
             if not check_name_is_well_formed:
                 e.filters.insert(len(e.filters), [func.lower(Name.name).op('~')(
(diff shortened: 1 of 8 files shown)


vysakh-menon-aot (approved):

severinbeauvais (approved):
The parts I understand look fine :)

Good job updating the unit tests!

PR #940, opened 2020-11-23
Selectively map NR attributes

bcdevlucas:
*Issue #, if available:*
/bcgov/entity#5171

*Description of changes:*


kialj876 (changes requested):
Looks good Lucas, but can you please add comments for why the pytests you added are skipped so that we know when/if we can unskip them later. Are there some existing pytests securing this work that are not skipped?

bcdevlucas:
@kialj876 
Pytests marked as skipped (in this PR) were just moved from another file - they need to be completed, but they're not directly related to this ticket. I just wanted to delete the files those tests were in. It's not new code... I was in the area so I cleared out stuff that has either been re-implemented or is no longer relevant.

PR #941, opened 2020-11-25
5687 stem distinctive

areyeslo:
*bcgov/entity#5687:*

*Description of changes:*
- Stemming distinctives and descriptives in dictionaries
- Search in synonyms for original word, if no result stemming word
- Adding gerund form for distinctives
- Deleted additional_dist_desc function, it is not longer needed after name examination session with Genevieve.
- Added test cases.

Regression Test:
image

Test 1 (plurals):
image

Test 2 (original distinctive - Mechanical in the same category as plumbing):
image


vysakh-menon-aot (approved):

vysakh-menon-aot (approved):

kialj876 (approved):

kialj876 (changes requested):
hey arturo can you fix the build? Looks like it's failing the GHA build due to the dataclasses version? ```dataclasses==0.8``` in requirements.txt should be something else I guess. I don't think you altered this but can you change it to whatever version you have running locally?

Other than that looks very good

PR #991, opened 2020-12-24
6031 add pending state flow

kialj876:
*Issue #, if available:*https://github.com/bcgov/entity/issues/6031

*Description of changes:*
- went through it with vysakh
- when payment is created for an nr sets state to pending payment until payment is completed


vysakh-menon-aot (approved):

api/namex/models/payment.py
```diff
@@ -62,6 +62,29 @@ def save_to_db(self):
...
+@event.listens_for(Payment, 'before_update')
+def update_nr_state(mapper, connection, target):
+    """Set the state of the NR based on payment_status_code."""
+    from namex.models import Request
+
+    payment = target
+    nr = Request.find_by_id(payment.nrId)
+    if nr:
+        # could not make this update properly via the model so used raw sql
+        if payment.payment_status_code == 'COMPLETED':
```
lekshmimallika-aot, line 75: For the scenarios like upgrade to priority, reapply, extension where a payment is involved is it ok to update the nr state like this? In other words, is there any other NR state for which a payment would come into picture?

thorwolpert:
This creates an error where a priority or extension gets hiccuped, but it blocks some other bad states, so we're rolling it in and correcting those asap

PR #999, opened 2021-01-05
Model changes for payment action

lekshmimallika-aot:
*Issue #, if available:*

*Description of changes:*
Updated payment model to have a new column for action
Updated the value of NR action from 'COMPLETE' to 'CREATE'
Updated test


api/namex/constants/__init__.py
```diff
@@ -392,7 +392,7 @@ class NameRequestActions(AbstractEnum):
     UPGRADE = 'UPGRADE'
     REAPPLY = 'REAPPLY'
     RECEIPT = 'RECEIPT'
-    COMPLETE = 'COMPLETE'
+    COMPLETE = 'CREATE'
```
kialj876, line 395: Is it worth changing the enum to CREATE as well so it doesn't confuse us later?

kialj876 (approved):

api/namex/constants/__init__.py
lekshmimallika-aot: Done

kialj876 (approved):

PR #1037, opened 2021-02-23
6335 send queue position and time estimate - one more fix

jordiwes:
*Issue # bcgov/entity#6335:

*Description of changes:*
fix for queue days (for UI


api/namex/resources/name_requests/name_request.py
```diff
@@ -54,10 +54,11 @@ def get(self, nr_id):
             if nr_model.stateCd == 'DRAFT':
                 oldest_draft = Request.get_oldest_draft().submittedDate
```
thorwolpert, line 55: i think this would throw an error if Request.get_oldest_draft() returned None.

jordiwes: Fixed as per comment

thorwolpert (approved):

PR #1047, opened 2021-03-18
6047 - handle unexpected status case

jordiwes:
*Issue #6047

*Description of changes:*

UI was receiving a response code of 0 from the api... I followed the path, and found this path where there was no exception thrown, but the api didn't receive an expected status code. In this case I print the status code, and return an error to the UI

### DIFF
--- a/api/namex/resources/payment/payment.py
+++ b/api/namex/resources/payment/payment.py
@@ -308,6 +308,12 @@ class CreateNameRequestPayment(AbstractNameRequestResource):
 
                 response = make_response(data, 201)
                 return response
+            # something went wrong with status code above
+            else:
+                # log actual status code
+                print('Error with status code. Actual status code: ' + payment_response.statusCode)
+                # return generic error status to the front end
+                return jsonify(message='Name Request {nr_id} encountered an error'.format(nr_id=nr_id)), 500
 
         except PaymentServiceError as err:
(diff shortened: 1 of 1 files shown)


api/namex/resources/payment/payment.py
```diff
@@ -308,6 +308,12 @@ def post(self, nr_id, payment_action=NameRequestActions.CREATE.value):
 
                 response = make_response(data, 201)
                 return response
+            # something went wrong with status code above
+            else:
+                # log actual status code
+                print('Error with status code. Actual status code: ' + payment_response.statusCode)
+                # return generic error status to the front end
+                return jsonify(message='Name Request {nr_id} encountered an error'.format(nr_id=nr_id)), 500
```
kialj876, line 316: Maybe thor will have thoughts on this, but in lear when we have a payment issue we usually return a 402. Probably good to be consistent here

kialj876 (approved):
Looks good - just one comment to consider

api/namex/resources/payment/payment.py
kialj876: we also usually use current_app.logger.debug() instead of print but doesn't effect anything in this case

jordiwes: Thanks! changes made

kialj876 (approved):

PR #1051, opened 2021-03-25
Openshift4 Migration and CICD setup.

pwei1018:
*Issue #, if available:*

*Description of changes:*


colin-api/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+ENV VCS_REF=${VCS_REF}
+ENV BUILD_DATE=${BUILD_DATE}
+
+LABEL org.label-schema.vcs-ref=${VCS_REF} \
+    org.label-schema.build-date=${BUILD_DATE}
+
+USER root
+
+# Create working directory
```
thorwolpert, line 14: colin will need the Oracle client files

colin-api/openshift/templates/bc.yaml
```diff
@@ -0,0 +1,127 @@
...
+        FROM docker-remote.artifacts.developer.gov.bc.ca/python:3.8.6-buster
+
+        ARG BUILD_DATE
+        ARG VCS_REF
+
+        LABEL build_date=$BUILD_DATE
+        LABEL vcs_ref=$VCS_REF
+
+        USER root
+
```
thorwolpert, line 48: this will need the oracle client libs

colin-api/openshift/templates/dc.yaml
```diff
@@ -0,0 +1,196 @@
...
+    value: dev
+    required: true
+
+  - name: ROLE
+    displayName: Role
+    description: Role
+    required: true
+    value: api
+
+  - name: NAMESPACE
```
thorwolpert, line 144: this will need the oracle envvars

colin-api/requirements.txt
```diff
@@ -6,6 +6,7 @@ Flask-SQLAlchemy
 Flask-Marshmallow
 Flask-OIDC
 flask-jwt-oidc
+Werkzeug==0.16.0
```
thorwolpert, line 9: can remove this, should be the newer one now

thorwolpert (commented):
couple of comments

colin-api/requirements.txt
thorwolpert: we'll open a techdeb tix for these

thorwolpert (approved):

colin-api/requirements.txt
pwei1018: I tried other version of Wekzeug but all get error. And also tried upgrade the flask-jwt-oidc to newest version. Same, not sure which library had issue.

colin-api/openshift/templates/dc.yaml
pwei1018: envvars will manage by 1password. So, I removed all envvars from dc template.

PR #1064, opened 2021-04-30
fixed pay issues

kialj876:

*Issue #, if available:* https://github.com/bcgov/entity/issues/7332

*Description of changes:*
- fixed pay logic around PAD and payment action mismatch

### DIFF
--- a/api/namex/models/payment.py
+++ b/api/namex/models/payment.py
@@ -2,6 +2,7 @@
 from enum import Enum
 from sqlalchemy import event
 
+from namex.constants import PaymentState
 from namex.models import State, db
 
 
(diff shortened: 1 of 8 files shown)


api/namex/models/payment.py
```diff
@@ -2,14 +2,15 @@
...
 
+from namex.constants import PaymentState
 from namex.models import State, db
 
 
 class Payment(db.Model):
 
     class PaymentActions(Enum):
         """Valid actions for a payment."""
-        COMPLETE='COMPLETE'
```
thorwolpert, line 12: this state is used by /services/namex-pay
i think CREATE was added, but maybe missed from this enum

thorwolpert (commented):
Payment.PaymentActions.COMPLETE is used by /services/namex-pay, so that would need to get fixed too.
The move to CREATE looks good otherwise.

api/namex/models/payment.py
kialj876: Yes I've included the namex-pay changes in this PR

thorwolpert: bah, missed the file path change. sweet.

thorwolpert (approved):

PR #1080, opened 2021-05-17
7433 Add submitted start/end date filter to requests search endpoint

argush3:
*Issue #:* /bcgov/entity#7433

*Description of changes:

* Added `submittedStartDate` and `submittedEndDate` search filters to `.../api/v1/requests` url.  
* Added test for `submittedStartDate` and `submittedEndDate` search filters
* Added files to ignore in `.gitignore`
* Added endpoint testing related utility functions under endpoint tests folder

### DIFF
--- a/api/namex/resources/requests.py
+++ b/api/namex/resources/requests.py
@@ -6,9 +6,7 @@ from flask import request, jsonify, g, current_app, get_flashed_messages
 from flask_restx import Namespace, Resource, fields, cors
 from flask_jwt_oidc import AuthError
 
-from pytz import timezone, UTC
-from dateutil import tz
-
+from namex.constants import DATE_TIME_FORMAT_SQL
 from namex.utils.logging import setup_logging
 
 from sqlalchemy.orm.exc import NoResultFound
@@ -297,7 +295,7 @@ class Requests(Resource):
                 '(now() at time zone \'utc\') - INTERVAL \'{hour_offset} HOURS\''.format(hour_offset=current_hour + 24 * 29)))
 
         if submittedInterval and (submittedStartDate or submittedEndDate):
-            return jsonify({"message": "submittedInterval cannot be used in conjuction with submittedStartDate and submittedEndDateTime"}), 400
+            return jsonify({"message": "submittedInterval cannot be used in conjuction with submittedStartDate and submittedEndDate"}), 400
 
         submittedStartDateTimeUtcObj = None
         submittedEndDateTimeUtcObj = None
@@ -306,7 +304,7 @@ class Requests(Resource):
             try:
                 submittedStartDateTimeUtcObj = convert_to_utc_min_date_time(submittedStartDate)
                 # convert date to format db expects
-                submittedStartDateTimeUtc = submittedStartDateTimeUtcObj.strftime('%Y-%m-%d %H:%M:%S%z')
+                submittedStartDateTimeUtc = submittedStartDateTimeUtcObj.strftime(DATE_TIME_FORMAT_SQL)
                 q = q.filter(RequestDAO.submittedDate >=
                              text('\'{submittedStartDateTimeUtc}\''
                               .format(submittedStartDateTimeUtc=submittedStartDateTimeUtc)))
@@ -318,7 +316,7 @@ class Requests(Resource):
             try:
                 submittedEndDateTimeUtcObj = convert_to_utc_max_date_time(submittedEndDate)
                 # convert date to format db expects
-                submittedEndDateTimeUtc = submittedEndDateTimeUtcObj.strftime('%Y-%m-%d %H:%M:%S%z')
+                submittedEndDateTimeUtc = submittedEndDateTimeUtcObj.strftime(DATE_TIME_FORMAT_SQL)
                 q = q.filter(RequestDAO.submittedDate <=
                              text('\'{submittedEndDateTimeUtc}\''
                                   .format(submittedEndDateTimeUtc=submittedEndDateTimeUtc)))
--- a/api/namex/resources/requests.py
+++ b/api/namex/resources/requests.py
@@ -6,6 +6,9 @@ from flask import request, jsonify, g, current_app, get_flashed_messages
 from flask_restx import Namespace, Resource, fields, cors
 from flask_jwt_oidc import AuthError
 
+from pytz import timezone, UTC
+from dateutil import tz
+
 from namex.utils.logging import setup_logging
 
 from sqlalchemy.orm.exc import NoResultFound
@@ -24,7 +27,7 @@ from namex.models import DecisionReason
 from namex.services import ServicesError, MessageServices, EventRecorder
 from namex.services.name_request.utils import check_ownership, get_or_create_user_by_jwt, valid_state_transition
 
-from namex.utils.common import convert_to_ascii
+from namex.utils.common import convert_to_ascii, convert_to_utc_date_time
 from namex.utils.auth import cors_preflight
 from namex.analytics import SolrQueries, RestrictedWords, VALID_ANALYSIS as ANALYTICS_VALID_ANALYSIS
 from namex.services.nro import NROServicesError
@@ -199,6 +202,8 @@ class Requests(Resource):
         submittedInterval = request.args.get('submittedInterval', None)
         lastUpdateInterval = request.args.get('lastUpdateInterval', None)
         current_hour = int(request.args.get('hour', 0))
+        submittedStartDateTime = request.args.get('submittedStartDateTime', None)
+        submittedEndDateTime = request.args.get('submittedEndDateTime', None)
 
         q = RequestDAO.query.filter()
         if queue:
@@ -223,15 +228,15 @@ class Requests(Resource):
         if compName:
             compName = compName.strip().replace(' ', '%')
             q = q.join(RequestDAO.names).filter(Name.name.ilike('%' + compName + '%'))
-        
+
         if firstName:
             firstName = firstName.strip().replace(' ', '%')
             q = q.join(RequestDAO.applicants).filter(Applicant.firstName.ilike('%' + firstName + '%'))
-        
+
         if lastName:
             lastName = lastName.strip().replace(' ', '%')
             q = q.join(RequestDAO.applicants).filter(Applicant.lastName.ilike('%' + lastName + '%'))
-        
+
         if consentOption == 'Received':
             q = q.filter(RequestDAO.consent_dt.isnot(None))
             print(q)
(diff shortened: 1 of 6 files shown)


api/tests/python/end_points/test_namex_search.py
```diff
@@ -346,3 +359,204 @@ def test_namex_search_last_name(client, jwt, app, search_name):
...
+
+    assert resp.get('nameRequests') and resp.get('response')
+    response_count = len(resp['nameRequests'][0])
+    assert response_count >= 0
+    assert response_count == expected_result_count
+
+# todo add test for submitted start and end date time
+
+@pytest.mark.parametrize('submitted_start_date_time, submitted_end_date_time', [
+    (escape_date_time('2021-05-11 00:00:00-07:00'), escape_date_time('2021-05-01 23:59:59-07:00')),
```
severinbeauvais, line 465: These date-time formats are inconsistent with our other UIs (eg, 2021-01-01T00:00:00.000000+00:00).

Also they are somewhat confusing ... if I understand correctly, the first example is "May 11, 2021 at 12:00:00 am in a timezone that is 7h behind UTC". But normally, date-times passed across the network are just plain UTC.

A couple of us chatted about this, this morning... I will let Cam discuss this with you further.

argush3: yes, talked to Cam a bit and will be refactoring such that FE won't need to provide anything but the date.  We discuss more after things are refactored.

severinbeauvais:
Does the API always return all results?

As you may know, the spec for the UI says it should display a message (and no results) if there are more than 1000 results... but it appears that the API has already searched for those and returned them, and the UI just throws them away, which is a waste of processing and network resources.

Would it be possible for the API to return a HTTP code (and no data) if there are more than 1000 results? Maybe one of these?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416

argush3:
> Does the API always return all results?
> 
> As you may know, the spec for the UI says it should display a message (and no results) if there are more than 1000 results... but it appears that the API has already searched for those and returned them, and the UI just throws them away, which is a waste of processing and network resources.
> 
> Would it be possible for the API to return a HTTP code (and no data) if there are more than 1000 results? Maybe one of these?
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416

I couldn't really modify existing behaviour too much as the examiners use a lot of the functionality in this endpoint.  To workaround this, the UI can request a page size of 1000 and as part of the response, if the `numFound` property is >= 1000, the UI can display the message defined in the requirements.  Is this good enough?  Or do we need to come up with another solution for the api?

``` 
{
  "nameRequests": [
    [],
    {}
  ],
  "response": {
    "numFound": 2394,
    "numPriorities": 0,
    "numUpdatedToday": 0,
    "order": "submittedDate desc NULLS LAST, stateCd desc NULLS LAST",
    "queue": null,
    "rows": 1000,
    "start": 0
  }
}

```

argush3:
@kialj876 do you see it being an issue doing what @severinbeauvais suggested?  i.e. returning a specific http status code with zero results when results exceed 1000?  i guess we could also add a specific flag to request params that would trigger that flow.

cameron-freshworks:
> > Does the API always return all results?
> > As you may know, the spec for the UI says it should display a message (and no results) if there are more than 1000 results... but it appears that the API has already searched for those and returned them, and the UI just throws them away, which is a waste of processing and network resources.
> > Would it be possible for the API to return a HTTP code (and no data) if there are more than 1000 results? Maybe one of these?
> > https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
> > https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413
> > https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416
> 
> I couldn't really modify existing behaviour too much as the examiners use a lot of the functionality in this endpoint. To workaround this, the UI can request a page size of 1000 and as part of the response, if the `numFound` property is >= 1000, the UI can display the message defined in the requirements. Is this good enough? Or do we need to come up with another solution for the api?
> 
> ```
> {
>   "nameRequests": [
>     [],
>     {}
>   ],
>   "response": {
>     "numFound": 2394,
>     "numPriorities": 0,
>     "numUpdatedToday": 0,
>     "order": "submittedDate desc NULLS LAST, stateCd desc NULLS LAST",
>     "queue": null,
>     "rows": 1000,
>     "start": 0
>   }
> }
> ```

This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results. 
@severinbeauvais Correct me if i'm wrong here tho!

argush3:
@severinbeauvais @cameron-freshworks @kialj876 does it work if I add a request param called `notifyOnlyResultSetMaxNum`.  If the count for the total result set exceeds whatever number is provided by the param, it will return a specific http status code with no results.

severinbeauvais:
> This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results.
> @severinbeauvais Correct me if i'm wrong here tho!

My concern is that we might pass a huge amount of data to the customer and then just throw it away. Some customers don't have a fast network and some customers pay a lot for their bandwidth, so this is a bad design.

Maybe this needs to be pushed back to Scott, ie, we would display the first 1000 anyway since we went through the trouble to find them and return them.

severinbeauvais:
> @severinbeauvais @cameron-freshworks @kialj876 does it work if I add a request param called `notifyOnlyResultSetMaxNum`. If the count for the total result set exceeds whatever number is provided by the param, it will return a specific http status code with no results.

What do other APIs do (in the real world)? I'm thinking a parameter like `countOnly` is standard but I haven't looked into this deeply.

api/namex/resources/requests.py
```diff
@@ -289,6 +296,40 @@ def get(*args, **kwargs):
             q = q.filter(RequestDAO.lastUpdate > text(
                 '(now() at time zone \'utc\') - INTERVAL \'{hour_offset} HOURS\''.format(hour_offset=current_hour + 24 * 29)))
 
+        if submittedInterval and (submittedStartDate or submittedEndDate):
+            return jsonify({"message": "submittedInterval cannot be used in conjuction with submittedStartDate and submittedEndDateTime"}), 400
```
severinbeauvais, line 300: Also "submittedEndDateTime" -> "submittedEndDate".

api/namex/utils/common.py
```diff
@@ -96,6 +99,24 @@ def convert_to_ascii(value):
         return value
 
 
+def convert_to_utc_min_date_time(date_str: str):
+    date_obj = datetime.strptime(date_str, '%Y-%m-%d').date()
+    min_time = time(hour=0, minute=0, second=0, microsecond=0)
+    return datetime.combine(date_obj, min_time, tzinfo=tz.UTC)
```
severinbeauvais, line 105: This changes the date string to, say "2021-04-17 00:00:00 Pacific" right? (ie, not 00:00 UTC)

api/namex/utils/common.py
```diff
@@ -96,6 +99,24 @@ def convert_to_ascii(value):
...
+def convert_to_utc_max_date_time(date_str: str):
+    date_obj = datetime.strptime(date_str, '%Y-%m-%d').date()
+    max_time = time(hour=23, minute=59, second=59, microsecond=999999)
+    return datetime.combine(date_obj, max_time, tzinfo=tz.UTC)
+
+
+# def convert_to_utc_date_time(date_time_str: str):
+#     """Convert datetime string with utc offset to datetime object"""
+#     date_time_obj = datetime.datetime.strptime(date_time_str, '%Y-%m-%d %H:%M:%S%z')
+#     return date_time_obj.astimezone(UTC)
```
severinbeauvais, line 117: Is this commented out because it's not used? What's the normal style here: leave in commented-out code, or delete it?

kialj876: > > This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results.
> > @severinbeauvais Correct me if i'm wrong here tho!
> 
> My concern is that we might pass a huge amount of data to the customer and then just throw it away. Some customers don't have a fast network and some customers pay a lot for their bandwidth, so this is a bad design.
> 
> Maybe this needs to be pushed back to Scott, ie, we would display the first 1000 anyway since we went through the trouble to find them and return them.
> 
> _Updated:_
> A fast response when there are too many results and you don't display anything is a nice side effect of this, yes.

Sorry I didn't see all these comments earlier. The API will only return the amount specified. So if you only want it to display 10 NRs you can do that by specifying rows=10 in the params. It will give you the total amount that matched in the reesponse (so you could show 10 out of 1 000,000 to the customer). It also supports paging with the 'start' param if that is also needed for you (so you could get 10 NRs starting from the 50th row for example)

kialj876:
> > This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results.
> > @severinbeauvais Correct me if i'm wrong here tho!
> 
> My concern is that we might pass a huge amount of data to the customer and then just throw it away. Some customers don't have a fast network and some customers pay a lot for their bandwidth, so this is a bad design.
> 
> Maybe this needs to be pushed back to Scott, ie, we would display the first 1000 anyway since we went through the trouble to find them and return them.
> 
> _Updated:_
> A fast response when there are too many results and you don't display anything is a nice side effect of this, yes.

Sorry I didn't see all these comments earlier. The API will only return the amount specified. So if you only want it to display 10 NRs you can do that by specifying rows=10 in the params. It will give you the total amount that matched in the reesponse (so you could show 10 out of 1 000,000 to the customer). It also supports paging with the 'start' param if that is also needed for you (so you could get 10 NRs starting from the 50th row for example)

kialj876:
> > > This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results.
> > > @severinbeauvais Correct me if i'm wrong here tho!
> > 
> > 
> > My concern is that we might pass a huge amount of data to the customer and then just throw it away. Some customers don't have a fast network and some customers pay a lot for their bandwidth, so this is a bad design.
> > Maybe this needs to be pushed back to Scott, ie, we would display the first 1000 anyway since we went through the trouble to find them and return them.
> > _Updated:_
> > A fast response when there are too many results and you don't display anything is a nice side effect of this, yes.
> 
> Sorry I didn't see all these comments earlier. The API will only return the amount specified. So if you only want it to display 10 NRs you can do that by specifying rows=10 in the params. It will give you the total amount that matched in the reesponse (so you could show 10 out of 1 000,000 to the customer). It also supports paging with the 'start' param if that is also needed for you (so you could get 10 NRs starting from the 50th row for example)

when I said 'total amount that matched in the reesponse' I meant it returns the total value. Not the all of the NRs

cameron-freshworks:
> > > > This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results.
> > > > @severinbeauvais Correct me if i'm wrong here tho!
> > > 
> > > 
> > > My concern is that we might pass a huge amount of data to the customer and then just throw it away. Some customers don't have a fast network and some customers pay a lot for their bandwidth, so this is a bad design.
> > > Maybe this needs to be pushed back to Scott, ie, we would display the first 1000 anyway since we went through the trouble to find them and return them.
> > > _Updated:_
> > > A fast response when there are too many results and you don't display anything is a nice side effect of this, yes.
> > 
> > 
> > Sorry I didn't see all these comments earlier. The API will only return the amount specified. So if you only want it to display 10 NRs you can do that by specifying rows=10 in the params. It will give you the total amount that matched in the reesponse (so you could show 10 out of 1 000,000 to the customer). It also supports paging with the 'start' param if that is also needed for you (so you could get 10 NRs starting from the 50th row for example)
> 
> when I said 'total amount that matched in the reesponse' I meant it returns the total value. Not all of the NRs

Just for context, the UI wants all 1000 of the Nr's if there is under 1000, but we don't present the results if there is over 1000. 

In this scenario, is the idea to make a request to see the number of matches and if under 1000 make another request to retrieve them?

severinbeauvais:
> The API will only return the amount specified. So if you only want it to display 10 NRs you can do that by specifying rows=10 in the params. It will give you the total amount that matched in the reesponse (so you could show 10 out of 1 000,000 to the customer). It also supports paging with the 'start' param if that is also needed for you (so you could get 10 NRs starting from the 50th row for example)

Wouldn't the UI have to request "1001" so that it could tell when there are more than 1000 results?

kialj876:
> @kialj876 do you see it being an issue doing what @severinbeauvais suggested? i.e. returning a specific http status code with zero results when results exceed 1000? i guess we could also add a specific flag to request params that would trigger that flow.

> > > > > This the current implementation you bet. I think Severins concern is when the results are large (in this case over 1000), the request takes some time to resolve and he would prefer a quick response with count prior to a response with all the results.
> > > > > @severinbeauvais Correct me if i'm wrong here tho!
> > > > 
> > > > 
> > > > My concern is that we might pass a huge amount of data to the customer and then just throw it away. Some customers don't have a fast network and some customers pay a lot for their bandwidth, so this is a bad design.
> > > > Maybe this needs to be pushed back to Scott, ie, we would display the first 1000 anyway since we went through the trouble to find them and return them.
> > > > _Updated:_
> > > > A fast response when there are too many results and you don't display anything is a nice side effect of this, yes.
> > > 
> > > 
> > > Sorry I didn't see all these comments earlier. The API will only return the amount specified. So if you only want it to display 10 NRs you can do that by specifying rows=10 in the params. It will give you the total amount that matched in the reesponse (so you could show 10 out of 1 000,000 to the customer). It also supports paging with the 'start' param if that is also needed for you (so you could get 10 NRs starting from the 50th row for example)
> > 
> > 
> > when I said 'total amount that matched in the reesponse' I meant it returns the total value. Not all of the NRs
> 
> Just for context, the UI wants all 1000 of the Nr's if there is under 1000, but we don't present the results if there is over 1000.
> 
> In this scenario, is the idea to make a request to see the number of matches and if under 1000 make another request to retrieve them?

Ah okay I think I'm confused by what you were discussing and the requirements. You only display the results if there's under 1000 and otherwise you just display nothing? Maybe we can have a chat or something. That will need it's own workflow that doesn't effect the examiner workflow

argush3:
@severinbeauvais @kialj876 @cameron-freshworks I think the code should be ready for review again now.  The timestamp and offset should no longer be required for the `submittedStartDate` and `submittedEndDate` request params now.

api/namex/resources/requests.py
argush3: Fixed this.  Thanks

severinbeauvais (approved):

api/namex/utils/common.py
argush3: This should be converting to UTC properly now with the latest updates.

image

api/namex/utils/common.py
argush3: Removed.

kialj876 (approved):
nice work!

PR #1107, opened 2021-06-09
6033 consume nr

vysakh-menon-aot:
*Issue* bcgov/entity#6033

*Description of changes:*

### DIFF
--- a/api/namex/resources/requests.py
+++ b/api/namex/resources/requests.py
@@ -531,6 +531,9 @@ class Request(Resource):
                             new_comment.nrId = nrd.id
 
                 ### END comments ###
+            elif consume := json_input.get('consume', None):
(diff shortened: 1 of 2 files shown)


api/namex/resources/requests.py
```diff
@@ -531,6 +531,9 @@ def patch(nr, *args, **kwargs):
                             new_comment.nrId = nrd.id
 
                 ### END comments ###
+            elif consume := json_input.get('consume', None):
```
vysakh-menon-aot, line 534: @thorwolpert do i need to do any state change for nr?

thorwolpert: no, that state doesn't change.
This will set the consumed info in the names DB, which then gets synced back to NameX.

thorwolpert (approved):

PR #1124, opened 2021-07-06
Added details section to pay-api payload

andrepestana-aot:
*Issue #:* /bcgov/entity#7253

*Description of changes:*
- Added details section to pay-api payload


kialj876 (commented):
Hey andre since you're in this code it might be good to add this in so we don't blow up every time relationships adds new stuff to the response: https://pypi.org/project/dataclasses-json/
add @dataclass_json(undefined=Undefined.EXCLUDE) to all those data classes
(sumesh sent me this a little while ago)

api/namex/resources/payment/utils.py
```diff
@@ -139,7 +141,52 @@ def merge_payment_request(nr_model, config=None):
...
+            'value': nr_model.nrNum
+        }
+    )
+    details.append(
+        {
+            'label': 'Name Choices:',
+            'value': ''
+        }
+    )
+    option1 = re.search('\|1(.*)1\|', nr_model.nameSearch)
```
kialj876, line 165: I don't think you should use the nameSearch column for this since that is specifically there for the advanced search and may change if we have time to try making it faster. Is there a reason you aren't using the names model? (i.e. nr_model.names.all() will give you all the names it has with choice/name/etc. info

kialj876 (changes requested):

api/namex/resources/payment/utils.py
andrepestana-aot: Yes, you're right. Fixed that. Thanks :)

andrepestana-aot:
> Hey andre since you're in this code it might be good to add this in so we don't blow up every time relationships adds new stuff to the response: https://pypi.org/project/dataclasses-json/
> add @dataclass_json(undefined=Undefined.EXCLUDE) to all those data classes
> (sumesh sent me this a little while ago)

Sure. I added it to the data classes.

api/namex/resources/payment/utils.py
```diff
@@ -139,7 +140,33 @@ def merge_payment_request(nr_model, config=None):
...
+            'value': nr_model.nrNum
+        }
+    )
+    details.append(
+        {
+            'label': 'Name Choices:',
+            'value': ''
+        }
+    )
+    for idx, name in enumerate(nr_model.names.all(), start=1):
```
kialj876, line 164: This looks good, but I think it's possible for the names to come in out of order. It's probably better to put the label as `name.choice + '.'` and then order them after if you need or order the names by the choice beforehand

kialj876 (commented):
Hey Andre, thanks for adding that stuff in. This looks great, just one comment to do with the name choices

api/namex/resources/payment/utils.py
andrepestana-aot: Yes, you're right. If we have the `choice` order number we must use it. :) thanks!

andrepestana-aot:
@kialj876 Can you please add @thorwolpert as a reviewer? ty

kialj876 (approved):

api/namex/services/payment/models/__init__.py
```diff
@@ -1,27 +1,32 @@
 from .abstract import Serializable
 from dataclasses import dataclass, field
+from dataclasses_json import dataclass_json, Undefined
```
thorwolpert, line 3: adding this package adds a 3rd json formatter to the toolset. should we add another?

thorwolpert (commented):
should we add a 3rd json formatter to the mix?

api/namex/services/payment/models/__init__.py
andrepestana-aot: I removed it.

thorwolpert (approved):

PR #1128, opened 2021-07-09
6968 Update wait time calculation logic to mirror NRO

argush3:
*Issue #:* /bcgov/entity#6968

*Description of changes:*

* Update wait time calculation logic in `statistics` endpoint to mirror NRO logic which is to use the oldest unexamined NR minus weekends


kialj876 (approved):
Nice work!

vysakh-menon-aot (approved):

PR #1138, opened 2021-07-17
7979 - Publish Cloud event when NR change state

andrepestana-aot:
*Issue #, if available: #7979*

*Description of changes:*

- Added CloudEventMessageService; 
- Added new configurations; 
- Added Request model event listener.


api/namex/services/cloud_event_messages/cloud_event_messages.py
```diff
@@ -0,0 +1,69 @@
...
+                                                utcfromtimestamp(time.time()).
+                                                replace(tzinfo=timezone.utc).
+                                                isoformat(),
+                                                identifier=nr_num,
+                                                json_data_body={
+                                                    'request': {
+                                                        'nrNum': nr_num,
+                                                        'oldNrNum': old_nr_num,
+                                                        'newState': state_cd,
+                                                        'previousState': old_state_cd
```
andrepestana-aot, line 53: Included new and old values as @sumesh-aot pointed out that the update from NRL to NR may be necessary:
https://github.com/bcgov/entity/pull/8003/files#r660844340

sumesh-aot: Do you store the account-id anywhere in names? To auto-affiliate the NR to the account, we will need to know the account to affiliate.

andrepestana-aot: Do you mean the account-id from the user who created the NR?

sumesh-aot: Yes, if it's an authenticated user.

sumesh-aot: Another option for us is to get the account information from Pay records for the NR.

api/namex/services/cloud_event_messages/cloud_event_messages.py
```diff
@@ -0,0 +1,69 @@
...
+                                                source=f'/requests/{nr_num}',
+                                                time=datetime.
+                                                utcfromtimestamp(time.time()).
+                                                replace(tzinfo=timezone.utc).
+                                                isoformat(),
+                                                identifier=nr_num,
+                                                json_data_body={
+                                                    'request': {
+                                                        'nrNum': nr_num,
+                                                        'oldNrNum': old_nr_num,
```
thorwolpert, line 51: Why is there an old NR num here?
The NRL is and only the NR is there when an invoice is created.
I'm not sure of the flow, but are we going to affiliate NRs that haven't been paid for?

thorwolpert (changes requested):
Might be worth copy/pasting in https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/queue.py

thorwolpert:
Looks like @vysakh-menon-aot already brought it in.
You'll need to rebase to pickup https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/services/queue.py

api/namex/services/cloud_event_messages/cloud_event_messages.py
andrepestana-aot: I can remove that. I'm just trying to satisfy what Sumesh pointed out: https://github.com/bcgov/namex/pull/1138#discussion_r671594620

andrepestana-aot: Please @sumesh-aot confirm we don't need the temporary NRL.

sumesh-aot: @andrepestana-aot @thorwolpert Not sure if NRL is being used now. 
Old and new identifier is needed if the name request uses NRL while creating an invoice, then only we can update the identifier to new identifier to properly show in receipt.

thorwolpert: An NR doesnt have both an NRL an a NR# so there's no way to map between the 2
It'll depend on if you affiliate an NR that hasn't been paid for yet.
we remove those after 24hrs, so it probably isn't worth it?

sumesh-aot: Affiliation will be based on the NR status change. So I don't think affiliation would happen before payment is received. But the need for old and new was for displaying in receipts and other payment records. 
- If the invoice was created with the NRL identifier, all the pay records would be having NRL. And if we need to update the NRL to NR, the event will need this info (similar to lear where the event contains temp registration number and permanent number).
- Also if we are planning to get the account information from pay records (one of my comment on this PR above) - we won't be able to get it if the records in pay stays NRL.

The second can be solved if the Names stores account-Id and send it in the event.

thorwolpert: I asked at standup, currently we wont be affiliating NRs that have been paid for, so we don't need to publish events for NRLs

thorwolpert: @sumesh-aot we're not there yet. it'll take more changes to support both NRL/NR# at the same time, as well as change the pay flow to be more like businesses (create the receipt with the temp reg, then change if paid/completed).

sumesh-aot: Yes, that aligns with my understanding. So do we get account-id in event (if NRL is being used for invoice creation) ?

sumesh-aot: aahh ok. Strike my above comment, it was a race condition :)

thorwolpert: Looks correct to me.
i don't think it should publish events for states prior to DRAFT (which is the state you enter once paid).

api/namex/services/cloud_event_messages/cloud_event_messages.py
andrepestana-aot: I see there's a user_id column for the `requests` and `events` tables but they are not populating with the user_id from the user who is creating the NR in namex database.

thorwolpert: it is if the user is logged in.

andrepestana-aot: Yes, it is possible to get the logged user from the request but namex-api is not saving the logged user. It saves name request account service userId in `requests` and `events` tables.

thorwolpert: ex. non-authenticated user submitted NR
```
   nr_num   |    state_cd     |           username
------------+-----------------+------------------------------
 NR 3640335 | DRAFT           | name_request_service_account
 NR 3640335 | PENDING_PAYMENT | name_request_service_account
 NR 3640335 | PENDING_PAYMENT | name_request_service_account
 NR 3640335 | DRAFT           | name_request_service_account
 NR 3640335 | INPROGRESS      | idir\kyee
 NR 3640335 | INPROGRESS      | idir\kyee
 NR 3640335 | INPROGRESS      | idir\kyee
 NR 3640335 | APPROVED        | idir\kyee
 NR 3640335 | APPROVED        | nro_service_account
 ```
 Looks like non staff are not captured via namerequest ... wtf

thorwolpert: in any case, first emitted message is in the DRAFT state, which is after payment.
so currently, PAY never has any NRLs, so there is no oldNrNum that is used.

andrepestana-aot: Yes, I fixed that. Only NRLs changes are not sending messages.

andrepestana-aot:
I'm creating a new PR because of the wrong merge to this branch

PR #1151, opened 2021-07-28
8222, 8413 - Added resubmit action

andrepestana-aot:
*Issue #, if available: #8222 and #8413*

*Description of changes:*

- Added RESUBMIT to actions for any NR that has 'EXPIRED', 'APPROVED' or 'CONDITIONAL, has past `expiration_date` and not consumed yet.
- Added RESUBMIT action as a valid request (it's the same as CREATE atm).
- Fixed `tests/python/end_points/payments/test_payments.py` and parameterized tests.
- Added more tests for temp NRs.
- Added a RESUBMIT test.

**The tests are still broken in GHA but it seems to be payment integration conf.**


kialj876 (commented):
Hey Andre, I don't think resubmit should be allowed for all states of the NR. Isn't resubmit only allowed after the NR is expired?

andrepestana-aot:
I'm changing it to DRAFT as it seems there is some things to be done for the RESUBMIT process.

severinbeauvais:
This message is to link this PR to tickets /bcgov/entity#8222 and /bcgov/entity#8413 .

api/config.py
```diff
@@ -74,12 +74,13 @@ class Config(object):
...
     DISABLE_NAMEREQUEST_NRO_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NRO_UPDATES', 0))
     DISABLE_NAMEREQUEST_SOLR_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_SOLR_UPDATES', 0))
+    DISABLE_NAMEREQUEST_NATS_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NATS_UPDATES', 0))
 
     # NATS
-    NATS_SERVERS = os.getenv('NATS_SERVERS')
-    NATS_CLIENT_NAME = os.getenv('NATS_CLIENT_NAME')
-    NATS_CLUSTER_ID = os.getenv('NATS_CLUSTER_ID')
-    NATS_QUEUE = os.getenv('NATS_QUEUE')
+    NATS_SERVERS = os.getenv('NATS_SERVERS', 'nats://localhost:4222')
```
severinbeauvais, line 80: Are you sure it's OK to commit any or all of these?

api/namex/constants/__init__.py
```diff
@@ -396,7 +462,8 @@ class NameRequestActions(AbstractEnum):
     INCORPORATE = 'INCORPORATE'
     # Payment related actions
     UPGRADE = 'UPGRADE'
-    REAPPLY = 'REAPPLY'
+    REAPPLY = 'REAPPLY' # FUTURE: replace by 'RENEW' - used when the NR is about to expire
+    RESUBMIT = 'RESUBMIT' # Used when the NR is already expired
```
severinbeauvais, line 466: Great comments 👍

api/namex/models/request.py
```diff
@@ -633,7 +633,7 @@ def on_insert_or_update_nr(mapper, connection, request):
        
        Temporary NRs (nrNum starting with 'NR L') are discarded.
     """
-    if not request.nrNum.startswith('NR L'):
+    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 and not request.nrNum.startswith('NR L'):
```
severinbeauvais, line 636: Is it OK that the fallback value for the "get" is 0? That would cause this condition to pass if you can't get the config, which doesn't seem safe...?

api/tests/python/end_points/name_requests/test_nr_actions.py
```diff
@@ -782,3 +793,78 @@ def test_draft_patch_resend(client, jwt, app):
 
     # Check NR number is the same because these are PATCH and call change_nr
     assert_field_is_mapped(draft_nr, patched_nr, 'nrNum')
+
+
+draft_input_fields = {
+	'applicants': [
```
severinbeauvais, line 799: Please check whether `applicants` is an array or simply an object. I saw this response to a "get NR" ...

image

api/config.py
```diff
@@ -74,12 +74,13 @@ class Config(object):
     # You can disable NRO updates for Name Requests by setting the variable in your .env / OpenShift configuration
     DISABLE_NAMEREQUEST_NRO_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NRO_UPDATES', 0))
     DISABLE_NAMEREQUEST_SOLR_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_SOLR_UPDATES', 0))
+    DISABLE_NAMEREQUEST_NATS_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NATS_UPDATES', 0))
```
severinbeauvais, line 77: What is this change for? I didn't see it mentioned in the PR description header.

api/config.py
andrepestana-aot: Yes, they are just default values to help other developers to configure it.

api/namex/models/request.py
andrepestana-aot: If it is 0, then it will be different of 1 and it will be true. I just want to disable it when someone intentionally add it to the environment vars. That will be helpful for testing locally without a NATS server. The same thing is done for SOLR and NRO updates.

api/config.py
andrepestana-aot: Please see: https://github.com/bcgov/namex/pull/1151#discussion_r683610058

api/tests/python/unit/test_setup_utils/build_nr.py
```diff
@@ -68,12 +69,12 @@ def build_nr(nr_state, data=None, test_names=None, generate_id_seq=None):
 
 def build_draft(data=None, test_names=None, generate_id_seq=None):
     try:
-        nr = RequestDAO()
+        nr = nr_json()
```
kialj876, line 72: Hey andre, what's the purpose of changing this method? The problem now is that all the other 'build_..' methods for the other states will create a different response type, which will make using *build_nr* very confusing later. Also, *build_nr* is a lot more useful for most tests if it returns the NR model and not just the json as you are doing here

kialj876 (changes requested):
This looks good! I think the test util fn needs to be changed back or altered though, but nice work

api/tests/python/unit/test_setup_utils/build_nr.py
andrepestana-aot: Yeah, ok... I don't think it's gonna work the way it was supposed to work anymore because of changes in the model but I'm gonna revert this and create my own NR json without using this function just to preserve it.

api/config.py
severinbeauvais: This conversation can be closed. (I do not have permission to do so.)

api/namex/constants/__init__.py
severinbeauvais: This conversation can be closed. (I do not have permission to do so.)

api/namex/models/request.py
severinbeauvais: This conversation can be closed. (I do not have permission to do so.)

api/config.py
severinbeauvais: This conversation can be closed. (I do not have permission to do so.)

api/tests/python/end_points/name_requests/test_nr_actions.py
andrepestana-aot: Both work when sending the data to create a new temp NR. The json representation sent back will send an object though:
https://github.com/bcgov/namex/blob/main/api/namex/models/request.py#L220

severinbeauvais: OK, thanks for checking!

This conversation can be closed. (I do not have permission to do so.)

severinbeauvais (approved):

api/tests/python/unit/test_setup_utils/build_nr.py
kialj876: okay sounds good

api/tests/python/end_points/name_requests/test_setup_utils/test_helpers.py
```diff
@@ -250,6 +252,27 @@ def post_test_nr(client, nr_data=None, nr_state=State.DRAFT):
         print(repr(err))
 
 
+@pytest.mark.skip
+def post_test_nr_json(client, nr_data=None):
```
kialj876, line 256: nice

kialj876 (approved):

PR #1157, opened 2021-08-06
8480 add staff auth header to fee calculation

vysakh-menon-aot:
*Issue:* bcgov/entity#8480

*Description of changes:*


severinbeauvais (approved):
Looks good, but what about the failing checks?

PR #1158, opened 2021-08-07
8488- Updates for new NATS messages and Resubmit payment action

andrepestana-aot:
*Issue:  bcgov/entity#8488*

*Description of changes:*

- Added config; 
- Initialized the queue;
- Updated the stan client_id everytime nats reconnects; 
- Initalized QueueService with a new event loop; 
- Added nest_asyncio to tests for the queue services work properly.

### DIFF
--- a/api/namex/services/queue.py
+++ b/api/namex/services/queue.py
@@ -74,12 +74,7 @@ class QueueService():
 
         default_stan_options = {
             'cluster_id': app.config.get('NATS_CLUSTER_ID'),
-            'client_id':
-            (self.name.
-             lower().
-             strip(string.whitespace)
-             ).translate({ord(c): '_' for c in string.punctuation})
-            + '_' + str(random.SystemRandom().getrandbits(0x58))
+            'client_id': self.get_client_id()
         }
         if not stan_options:
             stan_options = {}
@@ -107,7 +102,14 @@ class QueueService():
             if not ctx.nats.is_connected:
                 self.stan_options = {**self.stan_options, **{'nats': ctx.nats}}
                 await ctx.nats.connect(**self.nats_options)
+                # Update the client_id if nats is reconnecting
+                self.stan_options['client_id'] = self.get_client_id()
                 await ctx.stan.connect(**self.stan_options)
+
+    def get_client_id(self):
+        return (self.name.lower().strip(string.whitespace)).translate(
+                {ord(c): '_' for c in string.punctuation}
+               ) + '_' + str(random.SystemRandom().getrandbits(0x58))
 
     async def close(self):
         """Close the connections to the queue."""
--- a/api/namex/services/queue.py
+++ b/api/namex/services/queue.py
@@ -105,6 +105,7 @@ class QueueService():
                 # Update the client_id if nats is reconnecting
                 self.stan_options['client_id'] = self.get_client_id()
                 await ctx.stan.connect(**self.stan_options)
+                self.logger.warning('NATS has been connected with STAN client_id: %s', self.stan_options['client_id'])
 
     def get_client_id(self):
         return (self.name.lower().strip(string.whitespace)).translate(
--- a/services/namex-pay/config.py
+++ b/services/namex-pay/config.py
@@ -160,13 +160,6 @@ class TestConfig(Config):
     DISABLE_NAMEREQUEST_SOLR_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_SOLR_UPDATES', 0))
     DISABLE_NAMEREQUEST_NATS_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NATS_UPDATES', 1))
 
-    NATS_SERVERS = os.getenv('NATS_SERVERS', 'nats://localhost:4222')
-    NATS_CLIENT_NAME = os.getenv('NATS_CLIENT_NAME', 'namex.worker')
-    NATS_CLUSTER_ID = os.getenv('NATS_CLUSTER_ID', 'test-cluster')
-    NATS_QUEUE = os.getenv('NATS_QUEUE', 'namerequest-processor')
-    NATS_NR_STATE_SUBJECT = os.getenv('NATS_NR_STATE_SUBJECT', 'namex.event')
-    NATS_EMAILER_SUBJECT = os.getenv('NATS_EMAILER_SUBJECT', 'entity.email')
-
     # JWT OIDC settings
     # JWT_OIDC_TEST_MODE will set jwt_manager to use
     JWT_OIDC_TEST_MODE = True
--- a/services/namex-pay/config.py
+++ b/services/namex-pay/config.py
@@ -88,6 +88,14 @@ class Config(object):
     # You can disable NRO updates for Name Requests by setting the variable in your .env / OpenShift configuration
     DISABLE_NAMEREQUEST_NRO_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NRO_UPDATES', 0))
     DISABLE_NAMEREQUEST_SOLR_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_SOLR_UPDATES', 0))
+    DISABLE_NAMEREQUEST_NATS_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NATS_UPDATES', 0))
+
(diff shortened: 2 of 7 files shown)


severinbeauvais:
Could you please update the ticket number in the description to the "link format"? This automatically creates a link between this PR and the ticket. Thanks!

services/namex-pay/tests/unit/test_worker.py
```diff
@@ -62,7 +62,7 @@ def test_extract_payment_token():
          'is not',  # end_payment_has_value
          None  # error
          ),
-        ('already draft', Payment.PaymentActions.CREATE.value, State.DRAFT, State.DRAFT, 'N', 'N', datetime.utcnow(), 0, 'COMPLETED', 'COMPLETED', None, 'is', None),
+        ('already draft', Payment.PaymentActions.CREATE.value, State.DRAFT, State.DRAFT, 'N', 'N', datetime.utcnow(), 0, 'COMPLETED', 'COMPLETED', None, 'is not', None),
```
andrepestana-aot, line 65: There was a change in the business logic here.

services/namex-pay/src/namex_pay/worker.py
```diff
@@ -92,7 +93,7 @@ async def update_payment_record(payment: Payment) -> Optional[Payment]:
 
     payment_action = payment.payment_action
     nr = RequestDAO.find_by_id(payment.nrId)
-    if payment_action == Payment.PaymentActions.CREATE.value:  # pylint: disable=R1705
+    if payment_action in [Payment.PaymentActions.CREATE.value, Payment.PaymentActions.RESUBMIT.value]:  # pylint: disable=R1705
```
severinbeauvais, line 96: I know what's going because I'm reviewing this specific PR, but how about a comment here to explain to a future developer what's going on here?

severinbeauvais (approved):
LGTM but please wait for approvals from developers more familiar with NATS messaging.

kialj876 (approved):
Nice work, lgtm but I am also unfamiliar with the NATS setup now so this should wait for Thor I think

vysakh-menon-aot (approved):

andrepestana-aot:
@kialj876 as I added some env vars I think there's something to be configured to openshift as well?

api/namex/services/queue.py
```diff
@@ -107,8 +102,15 @@ async def connect(self):
             if not ctx.nats.is_connected:
                 self.stan_options = {**self.stan_options, **{'nats': ctx.nats}}
                 await ctx.nats.connect(**self.nats_options)
+                # Update the client_id if nats is reconnecting
+                self.stan_options['client_id'] = self.get_client_id()
```
thorwolpert, line 106: Why is this being changed?

services/namex-pay/config.py
```diff
@@ -150,6 +158,14 @@ class TestConfig(Config):
     # We can't run NRO locally for running our tests
     DISABLE_NAMEREQUEST_NRO_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NRO_UPDATES', 1))
     DISABLE_NAMEREQUEST_SOLR_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_SOLR_UPDATES', 0))
+    DISABLE_NAMEREQUEST_NATS_UPDATES = int(os.getenv('DISABLE_NAMEREQUEST_NATS_UPDATES', 1))
+
+    NATS_SERVERS = os.getenv('NATS_SERVERS', 'nats://localhost:4222')
```
thorwolpert, line 163: These are already here from ```Config``` use different values if you want it different in test.

services/namex-pay/src/namex_pay/worker.py
```diff
@@ -25,6 +25,8 @@
 the model to a standalone SQLAlchemy usage with an async engine would need
 to be pursued.
 """
+import asyncio
+import nest_asyncio
```
thorwolpert, line 29: What's the new library for?

services/namex-pay/src/namex_pay/worker.py
```diff
@@ -266,7 +270,8 @@ async def process_payment(pay_msg: dict, flask_app: Flask):
 FLASK_APP = Flask(__name__)
 FLASK_APP.config.from_object(APP_CONFIG)
 db.init_app(FLASK_APP)
-
+queue.init_app(FLASK_APP, asyncio.new_event_loop())
```
thorwolpert, line 273: why is this added?
the worker is already called from a loop in run.py

thorwolpert (commented):
i don't think adding the api q to the queue listener is correct.

api/namex/services/queue.py
andrepestana-aot: In namex-api it doesn't reconnect frequently but in namex-pay it reconnects often. In that part of the code it checks if nats is not connected and uses the same stan's client_id. When it happens it raises an error saying that the client_id is already connected.

services/namex-pay/config.py
andrepestana-aot: We need these keys for the QueueService from namex-api for all configs:
    NATS_SERVERS
    NATS_CLIENT_NAME
    NATS_CLUSTER_ID
    NATS_QUEUE
    NATS_NR_STATE_SUBJECT 
    NATS_EMAILER_SUBJECT

services/namex-pay/src/namex_pay/worker.py
andrepestana-aot: Quoting https://github.com/erdewit/nest_asyncio: 
`By design asyncio does not allow its event loop to be nested. This presents a practical problem: When in an environment where the event loop is already running it's impossible to run tasks and wait for the result. Trying to do so will give the error "RuntimeError: This event loop is already running".`

That's the only way I could make it work for namex-pay.

services/namex-pay/config.py
thorwolpert: right, and TestConfig already has those from Config

services/namex-pay/src/namex_pay/worker.py
thorwolpert: but why have you added the namex-api queue to the this listener, and in it's own loop.
solving a problem that maybe shouldn't have been created?

services/namex-pay/src/namex_pay/worker.py
andrepestana-aot: In run.py it runs QueueServiceManager. QueueService is a new queue manager added to namex-api and needs to be initialized in some way. I don't know if it's possible to change QueueService to work the same way as QueueServiceManager. What do you think?

andrepestana-aot:
> i don't think adding the api q to the queue listener is correct.

What do you mean? Having namex-api as dependency for namex-pay? What do you suggest?

services/namex-pay/src/namex_pay/worker.py
andrepestana-aot: Well, there's a problem. Maybe there's a better way to solve it:

Without adding the queue to namex-pay we got:
```
    raise err
  File "/home/andre/repos/namex/services/namex-pay/venv/lib/python3.8/site-packages/namex/services/queue.py", line 131, in publish_json_to_subject_sync
    self.loop.run_until_complete(self.async_publish_json(payload, subject))
AttributeError: 'NoneType' object has no attribute 'run_until_complete'

```

With:
`queue.init_app(FLASK_APP)`

```
  File "/home/andre/repos/namex/services/namex-pay/venv/lib/python3.8/site-packages/namex/services/queue.py", line 131, in publish_json_to_subject_sync
    self.loop.run_until_complete(self.async_publish_json(payload, subject))
  File "/usr/lib/python3.8/asyncio/base_events.py", line 592, in run_until_complete
    self._check_running()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 552, in _check_running
    raise RuntimeError('This event loop is already running')
RuntimeError: This event loop is already running
```

What do you suggest for this case?

services/namex-pay/config.py
andrepestana-aot: ok. Got it and removed.

api/namex/services/queue.py
thorwolpert: you've seen that in the logs?
I haven't seen it reconnecting unless the network has an error.
Over the past 14 days it has had 1 reconnect in Prod

thorwolpert (changes requested):
Just a couple of changes.
The other bits we'll refactor out with the upgrades.

PR #1161, opened 2021-08-12
8488 - raise exception when there is no payment record

andrepestana-aot:
*Issue #8488 :*

*Description of changes:*

- Raise an error when there is no record payment in namex db so the it can retry again.


thorwolpert:
this would need to be handled differently. please don't add this.

PR #1163, opened 2021-08-12
8554 fix transaction history for resubmit

vysakh-menon-aot:
*Issue:* bcgov/entity#8554

*Description of changes:*

### DIFF
--- a/api/namex/resources/events.py
+++ b/api/namex/resources/events.py
@@ -183,7 +183,7 @@ class Events(Resource):
                 Payment.PaymentActions.CREATE.value: 'Created NR',
                 Payment.PaymentActions.REAPPLY.value: 'Reapplied NR',
                 Payment.PaymentActions.UPGRADE.value: 'Upgraded Priority',
-                Payment.PaymentActions.RESUBMIT.value: 'Resubmited NR'
+                Payment.PaymentActions.RESUBMIT.value: 'Resubmitted NR'
             }
             for action in Payment.PaymentActions:
                 if action.value in e_dict['action']:
--- a/api/namex/resources/events.py
+++ b/api/namex/resources/events.py
@@ -182,7 +182,8 @@ class Events(Resource):
             payment_display = {
                 Payment.PaymentActions.CREATE.value: 'Created NR',
                 Payment.PaymentActions.REAPPLY.value: 'Reapplied NR',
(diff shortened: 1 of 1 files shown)


severinbeauvais (approved):

argush3 (approved):

andrepestana-aot:
Do we need this? I thought Severin would change RESUBMIT to NEW in the UI...

severinbeauvais:
> Do we need this? I thought Severin would change RESUBMIT to NEW in the UI...

Yes, this change is for the PAYMENT (not the REQUEST).

andrepestana-aot:
> > Do we need this? I thought Severin would change RESUBMIT to NEW in the UI...
> 
> Yes, this change is for the PAYMENT (not the REQUEST).

But as you're going to send NEW now we won't have any RESUBMIT anymore, right?

severinbeauvais:
> But as you're going to send NEW now we won't have any RESUBMIT payment anymore, right? How can we identify it is a 'new' NR and the payment should RESUBMIT?

The payment is created separately from the NR.

andrepestana-aot:
ok. So we're going to continue sending the request_action_cd as RESUBMIT to namex-api and namex-api needs to create a payment action as RESUBMIT? If so we'll need to change namex-api.

severinbeauvais:
> ok. So we're going to continue sending the request_action_cd as RESUBMIT to namex-api and namex-api needs to create a payment action as RESUBMIT? If so we'll need to change namex-api.

When the user clicks the confirm button...
1. first the NR object is created, with action=NEW - see https://github.com/bcgov/namerequest/blob/18e008a779170b05784faf3c8b8b3236124b9aa2/src/components/dialogs/resubmit.vue#L208
2. then the payment object is created, with action=RESUBMIT - see https://github.com/bcgov/namerequest/blob/18e008a779170b05784faf3c8b8b3236124b9aa2/src/components/dialogs/resubmit.vue#L227

andrepestana-aot:
ok. Yes. I think it's ok. Thanks!

severinbeauvais:
Just to close this off:

Andre, I think I understand your concern: when I asked yesterday about the impact to use NEW instead of RESUBMIT action, you thought both the NR action and payment action would be changed -- but only the NR action has been changed. Preliminary testing shows that both NR and payment are working correctly. Riyaz will help with more testing.

Yes, I also think it's OK :)

PR #1175, opened 2021-08-31
8721 NR State - CONSUMED, EXPIRED

vysakh-menon-aot:
*Issue:* bcgov/entity#8721

*Description of changes:*
  - New State - CONSUMED
  - Update NR state to EXPIRED


andrepestana-aot (commented):
Are the UIs prepared for this new state?

andrepestana-aot (approved):
Looks good!

argush3 (approved):

kialj876 (approved):
Looks good to me, but would be nice if you could add a couple tests around consumed NRs (I know most of the test suite fails, but we've been updating it in the areas we fix/add things)

vysakh-menon-aot:
> Looks good to me, but would be nice if you could add a couple tests around consumed NRs (I know most of the test suite fails, but we've been updating it in the areas we fix/add things)

@kialj876 I added some tests

api/namex/models/request.py
```diff
@@ -129,24 +129,11 @@ class Request(db.Model):
...
+        return self.stateCd == State.CONSUMED
 
     @property
     def is_expired(self):
-        todays_date = datetime.utcnow().date()
-        expiry_date = self.expirationDate.date() if self.expirationDate else None
-        if expiry_date:
-            return todays_date > expiry_date
-        return False
+        return self.stateCd == State.EXPIRED
```
thorwolpert, line 136: This is relying on the job to set EXPIRED, correct?
If so, would it not be safer to:
```suggestion
if State.EXPIRED or \
   self.expirationDate and datetime.utcnow().date() > self.expirationDate.date():
    return True
return False
```

thorwolpert (commented):
really good.
just one comment on expired state

api/namex/models/request.py
vysakh-menon-aot: I changed it here. I assume this won't happen since we have the job running. If the state did not change the UI's will be broken as its depending on EXPIRED status.

thorwolpert (approved):

PR #1193, opened 2021-11-19
8810_for_SP_GP_dont_call_queue_services

ozamani9:
*Issue #8810

*Description of changes:*
When nr is updated or inserted, a listener executes code to update queue services Do not call this function if the company entity_type_cd is FR (Sole prop) or GP (General Partnership)

### DIFF
--- a/api/namex/models/request.py
+++ b/api/namex/models/request.py
@@ -619,7 +619,9 @@ def on_insert_or_update_nr(mapper, connection, request):
 
        Temporary NRs (nrNum starting with 'NR L') are discarded.
     """
-    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 and not request.nrNum.startswith('NR L'):
+
+    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 and not request.nrNum.startswith('NR L') \
+        and request.entity_type_cd != 'FR' and request.entity_type_cd != 'GP':
         state_cd_history = get_history(request, 'stateCd')
         nr_num_history = get_history(request, 'nrNum')
         if len(nr_num_history.added) or len(state_cd_history.added):
--- a/api/namex/models/request.py
+++ b/api/namex/models/request.py
@@ -619,9 +619,9 @@ def on_insert_or_update_nr(mapper, connection, request):
 
        Temporary NRs (nrNum starting with 'NR L') are discarded.
     """
-
-    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 and not request.nrNum.startswith('NR L') \
-        and request.entity_type_cd != 'FR' and request.entity_type_cd != 'GP':
+    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 \
+       and not request.nrNum.startswith('NR L') \
+       and request.entity_type_cd not in ('FR', 'GP'):
         state_cd_history = get_history(request, 'stateCd')
         nr_num_history = get_history(request, 'nrNum')
         if len(nr_num_history.added) or len(state_cd_history.added):
--- a/api/tests/python/models/test_request.py
+++ b/api/tests/python/models/test_request.py
@@ -110,103 +110,36 @@ def test_has_consumed_name():
 
     assert nr.has_consumed_name is True
 
-def test_no_cloud_event_sent_for_approved_sp_gp_nrs():
-    """func description"""
-    from namex.models import Name, Request as RequestDAO, State
-    
-
-    with patch.dict(current_app.config, {'DISABLE_NAMEREQUEST_NATS_UPDATES': 0}):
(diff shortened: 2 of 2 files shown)


api/namex/models/request.py
```diff
@@ -619,7 +619,9 @@ def on_insert_or_update_nr(mapper, connection, request):
 
        Temporary NRs (nrNum starting with 'NR L') are discarded.
     """
-    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 and not request.nrNum.startswith('NR L'):
+
+    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 and not request.nrNum.startswith('NR L') \
+        and request.entity_type_cd != 'FR' and request.entity_type_cd != 'GP':
```
thorwolpert, line 624: You can use _in_ and break down the conditional a bit
```suggestion
    if current_app.config.get('DISABLE_NAMEREQUEST_NATS_UPDATES', 0) != 1 \
       and not request.nrNum.startswith('NR L') \
       and request.entity_type_cd not in ('FR', 'GP'):
```

api/tests/python/models/test_request.py
```diff
@@ -1,5 +1,8 @@
 from flask import jsonify
 from unittest import mock
+from unittest.mock import patch
+from namex.utils import queue_util
+from flask import current_app
 import pytest
```
thorwolpert, line 6: this will fail linting as it doesn't support PEP styling (base, then 3rd party, then local) imports
```suggestion
from unittest import mock
rom unittest.mock import patch

import pytest
from flask import current_app
from flask import jsonify

from namex.utils import queue_util
```
You can also put multiple imports on the same line if you want.

api/tests/python/models/test_request.py
```diff
@@ -107,6 +110,36 @@ def test_has_consumed_name():
 
     assert nr.has_consumed_name is True
 
+@pytest.mark.parametrize(['compName', 'compState', 'nrNum','nrEntity','assertValue'], [
+    ('TEST1', 'APPROVED','NR 0000001', 'FR', False),
```
thorwolpert, line 114: nice table test!
I'd only suggest giving a **test_name**, for each to give a hint as to what it is testing.

ozamani9: thanks Thor, will do that.   great suggestion.

api/namex/models/request.py
ozamani9: thank you for suggestion, changed, and repushed.

api/tests/python/models/test_request.py
ozamani9: dont fully understand this, but i implemented your suggestions, and repushed

api/tests/python/models/test_request.py
ozamani9: added testName, and repushed

api/tests/python/models/test_request.py
```diff
@@ -107,6 +112,36 @@ def test_has_consumed_name():
...
+    with patch.dict(current_app.config, {'DISABLE_NAMEREQUEST_NATS_UPDATES': 0}):
+        with patch.object(queue_util, 'send_name_request_state_msg', return_value=True) as mock_queue_util_send_msg_func:
+            name = Name()
+            name.choice = 1
+            name.name = compName
+            name.state = compState
+
+            nr = RequestDAO()
+            nr.nrNum= nrNum
+            if compState == 'APPROVED':
```
argush3, line 136: i think it's ok to just use the parameter value directly without doing the if/else for testing purposes.  simplify things.

ozamani9: i had trouble passing State.approved and State.conditional that's why i have the if condition

api/tests/python/models/test_request.py
thorwolpert: In our _lint_ setup, we use a package called _isort_
it has rules about how imports are laid out, sort order, etc. The pattern is sorted Top->Bottome, Left-Right alphabetically descending, and then grouped together:
import core packages from python
space
import 3rd party packages
space
import from this package

api/tests/python/models/test_request.py
argush3: those are just strings so you should be able to just pass them in as is.  you might have had troubles assigning `CONDITION` to `nr.stateCd`.  `CONDITIONAL` is a valid value.  `CONDITION` isn't.

argush3:
@ozamani9 I realized that filtering out queue messages for SP/GP NR entity types won't work.  This solution of updating the request model listener does stop the issue of conflicts showing up in names examination but introduces other issues.  Other queue consumers are listening for that event which is problematic.  I think the correct solution is to update `solr-names-updater` to not process NR state changes for NRs with SP/GP entity types.  cc: @thorwolpert

ozamani9:
After further investigation.  The code changes need to moved and tested in another area of the application.  solr-names-updater.

Closing Draft P.R.  will create another with the correct branch

PR #1196, opened 2021-11-26
10139 - Difference in expiry date between Namex and client results

andrepestana-aot:
*Issue #:* /bcgov/entity#10139

*Description of changes:*
- set expiration_date time to 11:59pm
- added some tests


api/namex/services/name_request/abstract_name_request.py
```diff
@@ -151,10 +151,10 @@ def generate_nr(cls):
         return nr_num
 
     @classmethod
-    def create_expiry_date(cls, start: datetime, expires_in_days: int, expiry_hour: int = 0, expiry_min: int = 1,
+    def create_expiry_date(cls, start: datetime, expires_in_days: int, expiry_hour: int = 23, expiry_min: int = 59,
                            tz: timezone = timezone('US/Pacific')) -> datetime:
-        """Create an expiry date in given days starting tomorrow."""
-        date = (start.astimezone(tz) + timedelta(days=expires_in_days + 1)) \
+        """Create an expiry date in given days and at 11:59pm."""
+        date = (start.astimezone(tz) + timedelta(days=expires_in_days)) \
```
severinbeauvais, line 157: I'm concerned that "start" is in one timezone (eg, Daylight time) but the target date is in another timezone (eg, Standard time). Adding a fixed number of days won't fix the timezone offset.

Or... does the Alchemy client correctly convert and save the date in the db ???

api/namex/models/request.py
```diff
@@ -133,8 +133,7 @@ def has_consumed_name(self):
 
     @property
     def is_expired(self):
-        return self.stateCd == State.EXPIRED or \
-            (self.expirationDate and datetime.utcnow().date() > self.expirationDate.date())
+        return self.stateCd == State.EXPIRED
```
severinbeauvais, line 136: Yes, as decided by Linda on Nov 24, that we would only look at the state.

The state is set by a daily job that we will now run right after 11:59 pm, so that NRs that expire at that time will have their new state without delay. (Currently there is a 7h delay between the expiry time and the state change, because the job is run at 7am.)

api/tests/python/services/name_request/test_abstract_name_request.py
```diff
@@ -0,0 +1,44 @@
...
+def test_create_expiry_date(input_datetime, expected_date):
+    """Test that create_expiry_date method returns a datetime at added X days and at 11:59pm.
+
+    It should not change the time (hour and minute) because of TimeZone offset changes through the year.
+    """
+    expiry_date = nr_svc.create_expiry_date(input_datetime, 56)
+    assert expiry_date
+    assert expiry_date.time().hour == 23
+    assert expiry_date.time().minute == 59
+    assert expiry_date.strftime('%z') in ['-0800', '-0700'] # should not be +0000 or any other
```
severinbeauvais, line 37: This should specifically look for a 8h offset during Standard time or a 7h offset during Daylight time. This can be a test parameter using known clock change values.

Ref: https://in-the-sky.org/timezone.php

api/tests/python/services/name_request/test_abstract_name_request.py
```diff
@@ -0,0 +1,44 @@
...
+    (datetime.datetime(2021, 3, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 5, 23)),
+    (datetime.datetime(2021, 4, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 6, 23)),
+    (datetime.datetime(2021, 5, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 7, 23)),
+    (datetime.datetime(2021, 6, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 8, 23)),
+    (datetime.datetime(2021, 7, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 9, 22)),
+    (datetime.datetime(2021, 8, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 10, 23)),
+    (datetime.datetime(2021, 9, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 11, 23)),
+    (datetime.datetime(2021, 10, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2021, 12, 23)),
+    (datetime.datetime(2021, 11, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2022, 1, 23)),
+    (datetime.datetime(2021, 12, 28, 14, 11, 00, tzinfo=pacific_tz), datetime.date(2022, 2, 22))
```
severinbeauvais, line 26: The hour and minute should be set to test edge cases.

The current values will work even if the timezone offset is incorrect (7h or 8h) because 11-7=4 is the same day as 11-8=3. (And vice versa for 11+7=18 and 11+8=19).

severinbeauvais: In other words, these should verify that a start date before the clock change should result in an expiry date with the correct time offset after the clock change.

api/tests/python/services/name_request/test_abstract_name_request.py
andrepestana-aot: yes, I know but we should keep track of the dates that it changes. The python libraries do their jobs and the date looks to be calculated correctly at this point.

severinbeauvais:
Please update bcgov/namex/api/namex/VERSION.py.

api/namex/services/name_request/abstract_name_request.py
severinbeauvais: outdated comment - closing

api/namex/services/name_request/name_request.py
```diff
@@ -144,7 +144,7 @@ def extend_expiry_date(cls, name_request, start_date=None, days=56):
         """
         Extends/sets the expiry date of an NR.
 
-        Default: extends the expiry date to 56 days from tomorrow's date at 00:01am.
+        Default: extends the expiry date to 56 days date at 11:59pm.
         """
         start_datetime = start_date if start_date else datetime.utcnow()
```
severinbeauvais, line 149: Does this need the same fix as above? I'm concerned about the possible assignment of a UTC date.

api/tests/python/services/name_request/test_abstract_name_request.py
```diff
@@ -0,0 +1,47 @@
...
+    Dates in 2021 between 2:00 a.m. on Sunday, March 14 to 2:00 a.m. on Sunday, November 7 should have their time
+    at 6:59am next day in UTC. Other dates should have their time at 7:59am next day in UTC.
+    """
+    expiry_date = nr_svc.create_expiry_date(input_datetime_utc, 56)
+    assert expiry_date
+    # Time in PST should always be 11:59pm
+    assert expiry_date.astimezone(pacific_tz).time().hour == 23
+    assert expiry_date.astimezone(pacific_tz).time().minute == 59
+
+    assert expiry_date.strftime('%z') in ['-0800', '-0700'] # should not be +0000 or any other
```
severinbeauvais, line 45: Is the purpose of this to verify the UTC offset in expiry_date? It this redundant, since you're checking the hour and minute in the next 2 lines?

If you think this is needed, I'd prefer to see the expected result as a test parameter, because checking for either -8 or -7 will not catch an off-by-one error.

api/tests/python/services/name_request/test_abstract_name_request.py
severinbeauvais: outdated comment - closing

api/tests/python/services/name_request/test_abstract_name_request.py
```diff
@@ -0,0 +1,47 @@
...
+    (datetime.datetime(2021, 7, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2021, 9, 22, 6, 59, 0, tzinfo=utc_tz)),
+    (datetime.datetime(2021, 8, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2021, 10, 23, 6, 59, 0, tzinfo=utc_tz)),
+    (datetime.datetime(2021, 9, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2021, 11, 23, 7, 59, 0, tzinfo=utc_tz)),
+    (datetime.datetime(2021, 10, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2021, 12, 23, 7, 59, 0, tzinfo=utc_tz)),
+    (datetime.datetime(2021, 11, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2022, 1, 23, 7, 59, 0, tzinfo=utc_tz)),
+    (datetime.datetime(2021, 12, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2022, 2, 22, 7, 59, 0, tzinfo=utc_tz))
+])
+def test_create_expiry_date(input_datetime_utc, expected_date_utc):
+    """
+    Test that create_expiry_date method returns a datetime at added X days and at 11:59pm.
```
severinbeauvais, line 32: Please change to "at 11:59 pm Pacific". (This is so confusing, I just want to be super clear.)

api/namex/services/name_request/abstract_name_request.py
```diff
@@ -151,13 +151,25 @@ def generate_nr(cls):
         return nr_num
 
     @classmethod
-    def create_expiry_date(cls, start: datetime, expires_in_days: int, expiry_hour: int = 0, expiry_min: int = 1,
+    def create_expiry_date(cls, start: datetime, expires_in_days: int, expiry_hour: int = 23, expiry_min: int = 59,
```
severinbeauvais, line 154: If there is no calling function that sets expiry_hour/min, I would be prefer these parameters to be removed -- it may prevent a future logic bug.

(Same with timezone: if it is always UTC, and should always be UTC, remove the param.)

severinbeauvais: corrected my comment above ^^^

api/tests/python/services/name_request/test_abstract_name_request.py
```diff
@@ -0,0 +1,47 @@
...
+from namex.models import Request as RequestDAO
+
+nr_svc = NameRequestService()
+pacific_tz = gettz('US/Pacific')
+utc_tz = gettz('UTC')
+
+
+@pytest.mark.parametrize('input_datetime_utc,expected_date_utc', [
+    (datetime.datetime(2021, 1, 28, 14, 11, 00, tzinfo=utc_tz), datetime.datetime(2021, 3, 26, 6, 59, 0, tzinfo=utc_tz)),
+    (datetime.datetime(2021, 1, 28, 11, 11, 00, tzinfo=utc_tz), datetime.datetime(2021, 3, 25, 6, 59, 0, tzinfo=utc_tz)),
```
severinbeauvais, line 17: I think it would be better to test edge cases instead of 14:11:00 and 11:11:00, since these won't catch off-by-7 or off-by-8 errors.

In fact, I don't understand why lines 16 and 17 don't resolve to the same date ???

api/namex/services/name_request/abstract_name_request.py
severinbeauvais: fixed

api/tests/python/services/name_request/test_abstract_name_request.py
andrepestana-aot: >In fact, I don't understand why lines 16 and 17 don't resolve to the same date ???

Yes, you're right. As I wasn't checking any dates the test was ok but yes, there was an issue because I was adding days with UTC. I fixed that and added tests to check the edge cases. THANKS!

api/namex/services/name_request/abstract_name_request.py
```diff
@@ -151,13 +151,26 @@ def generate_nr(cls):
...
     @classmethod
-    def create_expiry_date(cls, start: datetime, expires_in_days: int, expiry_hour: int = 0, expiry_min: int = 1,
-                           tz: timezone = timezone('US/Pacific')) -> datetime:
-        """Create an expiry date in given days starting tomorrow."""
-        date = (start.astimezone(tz) + timedelta(days=expires_in_days + 1)) \
-            .replace(hour=expiry_hour, minute=expiry_min, second=0, microsecond=0)
-
-        return date
+    def create_expiry_date(cls, start: datetime, expires_in_days: int):
+        """Create an expiry date in given days and at 11:59pm.
```
severinbeauvais, line 155: Please change to "at 11:59 pm Pacific" here also :)

api/tests/python/services/name_request/test_abstract_name_request.py
severinbeauvais: The new tests look great !!!

api/tests/python/services/name_request/test_abstract_name_request.py
severinbeauvais: Thanks. It's actually more readable / easier to understand and verify now, too!

api/namex/services/name_request/abstract_name_request.py
```diff
@@ -151,13 +151,26 @@ def generate_nr(cls):
...
+        expiry_hour = 23
+        expiry_min = 59
+
+        # convert the input date to PST and removes tzinfo:
+        naive_date_pst = start.astimezone(pacific_tz).replace(tzinfo=None)
+        # calculate the new day in PST
+        expiry_date_pst = (naive_date_pst + timedelta(days=expires_in_days))
+        # make it localized back to PST 
+        expiry_date_pst_localized = pacific_tz.localize(expiry_date_pst)
+        # set the time to 11:59pm in PST
```
severinbeauvais, line 170: This might now be "PST" (Pacific Standard Time = winter time). It might be PDT.
I suggest you just change this to "Pacific".
Also lines above.

api/tests/python/services/name_request/test_abstract_name_request.py
severinbeauvais: fixed, thanks

api/tests/python/services/name_request/test_abstract_name_request.py
andrepestana-aot: I changed it to specific offset.

api/namex/services/name_request/name_request.py
andrepestana-aot: I don't think we need to change it here. What are you comparing with? (above?)

api/tests/python/services/name_request/test_abstract_name_request.py
andrepestana-aot: Added edge cases.

api/namex/services/name_request/name_request.py
severinbeauvais: It liks like it can set start_datetime as "now" in UTC. Should that be "now" local?

andrepestana-aot: As it is timezone aware it is ok. If you create 2 datetimes: one with Pacifc time and another in UTC with the same datetime, both when converted to the same timezone should have the same values.

severinbeauvais: OK

severinbeauvais (approved):
Please make sure you get approval from the real namex experts :)

api/namex/resources/payment/payment.py
```diff
@@ -388,7 +388,7 @@ def post(self, nr_id, payment_action=NameRequestActions.CREATE.value):
                                 msg = f'Extend NR for payment.id={payment.id} nr_model.state{nr_model.stateCd}, nr_model.expires:{nr_model.expirationDate}'
                                 current_app.logger.debug(msg)
 
-                            nr_model.expirationDate = nr_model.expirationDate + timedelta(days=NAME_REQUEST_LIFESPAN_DAYS)
+                            nr_model.expirationDate = nr_svc.create_expiry_date(nr_model.expirationDate, NAME_REQUEST_LIFESPAN_DAYS)
```
severinbeauvais, line 391: That was easy :)

Now how does this get tested?

thorwolpert (approved):
the tests look ok

kialj876 (approved):

api/namex/resources/payment/payment.py
andrepestana-aot: Sorry, I tried to create a test for the request but the tests in namex are really complicate to me and I got blocked by authentication.

argush3 (approved):

PR #1199, opened 2021-12-01
8810_sp_gp_nr_approvals_showing_as_conflicts

ozamani9:
*Issue #8810 

*Description of changes:*

When nr is updated or inserted, a listener executes code to update queue services Do not call this function if the company entity_type_cd is FR (Sole prop) or GP (General Partnership)


argush3 (approved):
looks good!

PR #1205, opened 2021-12-20
7087 restore expire 1year 56days

ozamani9:
*Issue /bcgov/entity#7087

*Description of changes:*

Before creating or updating expiration date call a function that determines if the number of days to apply for expiration date is either 421 (REH, REN) or 56 (all other types)


api/namex/constants/__init__.py
```diff
@@ -669,3 +669,7 @@ class EventUserId(AbstractEnum):
 class RequestPriority(AbstractEnum):
     Y = 'Y'
     N = 'N'
+
+class ExpiryDays(AbstractEnum):
```
thorwolpert, line 673: if you change to ```(str, AbstractEnum)```
then you can compare to a string value, makes it a little easier to use.

api/namex/services/name_request/name_request.py
```diff
@@ -140,24 +140,38 @@ def update_request_submit_count(cls, name_request):
...
-                expires_in_days=days
+                expires_in_days=expiry_days
             )
         except Exception as err:
             raise ExtendExpiryDateError(err)
 
         return name_request
 
-    def map_request_data(self, name_request, map_attrs=False):
+    def map_request_data(cls, self, name_request, map_attrs=False):
```
thorwolpert, line 174: This needs to be decorated as a class method
```suggestion
    @classmethod
    def map_request_data(cls, self, name_request, map_attrs=False):
```
You can accept the suggestion here and it will change the code.

thorwolpert (commented):
lgtm, a couple of suggested changes.

api/namex/constants/__init__.py
ozamani9: updated this, and i also updated version file (i forget that needs to be done also)

thorwolpert (approved):
good to go!

eve-git (approved):
LGTM

PR #1206, opened 2022-01-05
9499 update my business registry target table

vysakh-menon-aot:
*Issue:* /bcgov/entity#9499

*Description of changes:*


argush3 (approved):

api/namex/services/lookup/name_request_filing_actions.py
```diff
@@ -63,8 +63,8 @@ def teardown(self, exception):
...
         ('RLC', 'LLC', 'Extraprovincial Limited Liability Company', 'Extraprovincial Reinstatement', 'static', None, 'https://www2.gov.bc.ca/gov/content/employment-business/business/managing-a-business/permits-licences/businesses-incorporated-companies/forms-corporate-registry#information-packages', None),
         ('AL', 'LLC', 'Extraprovincial Limited Liability Company', 'Extraprovincial Assumed Name', 'static', None, 'https://www2.gov.bc.ca/assets/gov/employment-business-and-economic-development/business-management/permits-licences-and-registration/registries-forms/form_37_xco_-_name_change.pdf', None),
-        ('FR', 'GP', 'General Partnership/Sole Proprietorship/DBA', 'Registration', 'onestop', None, None, None),
-        ('CFR', 'GP', 'General Partnership/Sole Proprietorship/DBA', 'Change of Name', 'onestop', None, None, None),
+        ('FR', 'GP', 'General Partnership/Sole Proprietorship/DBA', 'Registration', 'lear', None, None, None),
```
vysakh-menon-aot, line 66: @thorwolpert we might need to change this mapping to support the new design of SP/GP registration. As per the design the lear should be able to identify whether its a General Partnership or Sole Proprietorship or DBA. While making the change here we need to change `request_type_mapping` for the same

thorwolpert: No, that value is used by the auth dashboard to know where to route to.
It should be ok to change the value here, as the auth ui will pick it up.

thorwolpert (approved):

PR #1213, opened 2022-02-08
Add error tracking when got create payment exception

pwei1018:
*Issue #, if available:* https://github.com/bcgov/entity/issues/10052

*Description of changes:*
The issue is the payment record didn't get create after submit the payment from NRA due to some reasons. 
1. I added an event record when exception happen, also switch debug to error logging;
2. Create payment record before submit payment;


pwei1018:
This change will not fix the issue. It just add more visibility to the issue.

ozamani9 (approved):
Looks good.  If you have to make any other code changes, perhaps you can update version number in    /namex/api/namex/VERSION.py

pwei1018:
> Looks good. If you have to make any other code changes, perhaps you can update version number in /namex/api/namex/VERSION.py

done.

PR #1251, opened 2022-06-14
fix python prorobuf error

eve-git:
*Issue #, if available:*
12489

*Description of changes:*
Python has a [bug](https://github.com/protocolbuffers/protobuf/issues/10051) in release 4.21.0. Add “protobuf==3.20.1” to requirments to a temporary solution. Once the bug is fixed, remove the change.


eve-git:
@pwei1018 how to know when the python bug is fixed? Is someone responsible to remove the line in the requirement when python bug fixed?

PR #1317, opened 2022-12-08
user idp_userid as unique identifier

bolyachevets:
*Issue #, if available:*
bcgov/entity#14596

*Description of changes:*
after keycloak migration to gold, sub from jwt token (i.e. keycloak guid) will be changed for users, as it is tied to a specific IDP instance, and idp guid should be used instead to uniquely identify users

### DIFF
--- a/api/namex/services/name_request/utils.py
+++ b/api/namex/services/name_request/utils.py
@@ -123,7 +123,7 @@ def check_ownership(nrd, user):
 def get_or_create_user_by_jwt(jwt_oidc_token):
     # GET existing or CREATE new user based on the JWT info
     try:
-        user = User.find_by_jwt_idp_userid(jwt_oidc_token)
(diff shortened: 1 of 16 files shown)


api/namex/services/name_request/utils.py
```diff
@@ -123,7 +123,7 @@ def check_ownership(nrd, user):
 def get_or_create_user_by_jwt(jwt_oidc_token):
     # GET existing or CREATE new user based on the JWT info
     try:
-        user = User.find_by_jwtToken(jwt_oidc_token)
```
thorwolpert, line 126: This lookup is made in a few other places.
Will you be changing those as well?
I'm not sure if the tests still work for this, but if they do, it would be good to update them as well

thorwolpert (approved):
thanks for covering all those changes and tests!

pwei1018 (approved):

pwei1018 (approved):

PR #1348, opened 2023-02-09
check if nr was found in db

bolyachevets:
*Issue #, [Namex API: shouldn't crash (500) on invalid Corp Num](https://github.com/bcgov/entity/issues/15026)
[entity #15026](https://github.com/bcgov/entity/issues/15026)

*Description of changes:*

I traced the issue to colin-api, where in the logs you would get:

```
incorp_reg_addr_id = incorp_addr_ids[0][0]
IndexError: list index out of range
```
Looking at this I have more questions than answers. I could observe the 500 error only in dev environment and not test/prod, where in the latter we would be getting a 404 instead.

This would fix the 500 in dev, but is this the desired behavior (rather than exception message.. although I do feel uneasy using exceptions for index out of range errors)


eve-git (approved):
Looks good to me. 
Can you put the SRE ticket number in the description in this ticket?

ozamani9gh (approved):
Looks good to me

PR #1366, opened 2023-02-27
15175 - Changes for search for SBC-AUTH, other various unit test fixes

seeker25:
*Issue #, if available:* 
https://github.com/bcgov/entity/issues/15175

*Description of changes:*
Optimized search route for auth to call, without disturbing the operations of namex.


api/tests/python/end_points/test_namex_search.py
```diff
@@ -603,58 +603,30 @@ def test_namex_search_submitted_start_and_end_date_invalid_date_format(client,
...
-    assert rv.status_code
-    assert rv.status_code == 200
-    assert rv.data
-    resp = json.loads(rv.data.decode('utf-8'))
-    assert len(resp) == 5
-
-
 @pytest.mark.parametrize('test_name, identifiers, total_results', [
-    ('Search for NRs by identifier', ['NR 0', 'NR 1', 'NR 2', 'NR 3'], 4),
+    ('Search for NRs by identifier', ['NR 0', 'NR 1', 'NR 2', 'NR 3', 'NR 4'], 5),
```
kialj876, line 607: better to only query 4 of the 5 created for the test I think

api/tests/python/end_points/test_namex_search.py
```diff
@@ -603,58 +603,30 @@ def test_namex_search_submitted_start_and_end_date_invalid_date_format(client,
...
         [{'name': 'test1', 'state': 'NE', 'choice': 1}],
         [{'name': 'test 1', 'state': 'NE', 'choice': 1}],
         [{'name': 'test tester 1', 'state': 'NE', 'choice': 1}],
         [{'name': 'testing tester 1', 'state': 'NE', 'choice': 1}],
         [{'name': 'test tester 1', 'state': 'NE', 'choice': 1}]
     ]
-    generate_nrs(5, [], names, [])
+    base_nrs = generate_nrs(5, [], names, [])
+    for nr in base_nrs:
+        nr.save_to_db()
```
kialj876, line 621: don't need this they already get saved

api/tests/python/end_points/test_namex_search.py
```diff
@@ -603,63 +605,53 @@ def test_namex_search_submitted_start_and_end_date_invalid_date_format(client,
...
+    base_names = [
         [{'name': 'test1', 'state': 'NE', 'choice': 1}],
         [{'name': 'test 1', 'state': 'NE', 'choice': 1}],
         [{'name': 'test tester 1', 'state': 'NE', 'choice': 1}],
         [{'name': 'testing tester 1', 'state': 'NE', 'choice': 1}],
         [{'name': 'test tester 1', 'state': 'NE', 'choice': 1}]
     ]
-    generate_nrs(5, [], names, [])
+    base_applicants = [
+        create_applicant('1', 'ted', email_address='ted', phone_number='1'),
```
kialj876, line 624: can you add this to the generate_nrs fn with a default set? That way all the tests that use it will have one. Would be good to have a default name created too so that the tests don't need to set names each time unless they need to test specific values

seeker25:
I tried creating an NR, seemed to work:
image

seeker25:
1000 NR's, took about 2.62 seconds:

image


This will need through testing after merging, I might need some help testing all the functionality in namex.

kialj876 (approved):
Nice work!

PR #1417, opened 2023-09-15
17505 Create a loop to attempt retrieval of payment records from the payment queue callback up to a maximum of five times

eve-git:
*Issue #, if available:*
[17505](https://github.com/bcgov/entity/issues/17505)

*Description of changes:*
Create a loop to attempt retrieval of payment records from the payment queue callback up to a maximum of five times


services/namex-pay/src/namex_pay/worker.py
seeker25, line 1: We should probably throw an exception on line 259, that way the queue will reprocess the message if the retry fails.

seeker25 (commented):
comment above

services/namex-pay/src/namex_pay/worker.py
eve-git: @seeker25 please review

seeker25: Looks good, you might not need the retry loop with the exception being thrown, but I think this is ok for now.

seeker25 (approved):

bolyachevets (approved):

services/namex-pay/requirements.txt
ozamani9gh, line 1: why are all these packages changed for this request

seeker25: because when you run make setup and scroll up you'll see alot of red errors with incompatible packages - these are added to ensure the service picks up the right requirements

ozamani9gh (approved):
Thanks Eve, looks good

PR #1437, opened 2023-09-27
#17144 Duplicate name request issue.

Rajandeep98:
…ta, created exception to catch error for duplicates

*Issue #, if available:* https://github.com/bcgov/entity/issues/17144

*Description of changes:*
updated exception


api/namex/services/name_request/exceptions.py
```diff
@@ -128,3 +128,8 @@ def __init__(self, wrapped_err=None, message="Error getting user id."):
 class VirtualWordConditionServiceError(NameRequestException):
     def __init__(self, wrapped_err=None, message="Error initializing VirtualWordCondition Service."):
         super().__init__(wrapped_err, message)
+
+###################rajan exception raising for existing by same name and by same user
```
eve-git, line 132: Remove your name from the comment?

api/namex/services/name_request/exceptions.py
```diff
@@ -128,3 +128,8 @@ def __init__(self, wrapped_err=None, message="Error getting user id."):
...
+class NameRequestIsAlreadySubmittedError(NameRequestException):
+    def __init__(self, wrapped_err=None, message="The request with same name is already in process."):
```
eve-git, line 134: I think "in process" here means starting examining. It might be more precise to use 'submitted' instead..

PR #1438, opened 2023-09-28
documenting oracle trigger update

kialj876:
*Issue #, if available:* https://github.com/bcgov/entity/issues/15978

*Description of changes:*


kialj876:
This is just documentation of what was updated inside CDEV/TST and what will be in CPRD (doesn't get deployed anywhere)

PR #1484, opened 2023-11-09
18515_Unblock_duplicate_after_5min_interval

Rajandeep98:
*Issue #18515 - URGENT HOTFIX - PENDING_PAYMENT STATUS INCORRECT - Submitting intentional duplicate names blocked
https://github.com/bcgov/entity/issues/18515

*Description of changes:*
Updating the interval to prevent accidental duplicates and allowing to create intentional duplicate name request after 5 minutes if user needs it.


api/namex/models/request.py
```diff
@@ -331,11 +331,14 @@ def find_existing_name_by_user(cls, user_name_search_string, email):
...
         Gets requests submited by user with given name choice in state draft
         """
+        current_time = datetime.utcnow()
         existing_nr = db.session.query(Request). \
             join(Applicant, and_(Applicant.nrId == Request.id)). \
             filter(
                 Applicant.emailAddress == email,
                 (Request.stateCd == 'DRAFT') | (Request.stateCd == 'PENDING_PAYMENT') ,
+                #check if the request is being made within 5 minutes, if time >= 5 minutes, it will let the user create a new request with same name otherwise it will give the error message to prevent accidental duplicates and accidental double payments for them.
+                Request.submittedDate >= current_time - timedelta(minutes=5),
```
bolyachevets, line 341: i am curious, why is it 5 minutes exactly and not 1 or 10?

api/namex/models/request.py
```diff
@@ -331,11 +331,14 @@ def find_existing_name_by_user(cls, user_name_search_string, email):
...
                 (Request.nameSearch == ('('+user_name_search_string+')')) | ( Request.nameSearch == user_name_search_string )). \
             one_or_none()
```
bolyachevets, line 343: also, this method throws an exception if more than one row is returned: https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one_or_none - maybe i don't understand the logic, but if we actually get multiple rows, would not this result in an unhandled exception

api/namex/models/request.py
ozamani9gh: they want to allow for users that error out to resend the exact same name, after a 5 minute window.  Waiting on Joshua to approve this 5 minutes.   we may lower the time limit.  we just dont want users to DOUBLE Process the exact same Name Request

ozamani9gh: my question is current_time is being set in the same routine that compares the submitted_date.  i think this will always be false.

I think we need an array stored somewhere

submittedDate, when Request to create a DRAFT NR is processed
Applicant_Name
Name Choice 1
Name Choice 2
Name Choice 3.

and a timer after 5 minutes to reset these stored values.

your code should compare the request info with what is stored.

if you are achieving this in the above code, awesome, but i just dont see that.

api/namex/models/request.py
ozamani9gh: yes please fix that exception, Rajan

ozamani9gh (changes requested):
please look at my comments

api/namex/models/request.py
Rajandeep98: updated code to check on only for pending payment state if request stuck for payment about 5 mins or more then user can send in another request, once nr is fully completed with payment then system will not allow duplicates not even intentional ones.

ozamani9gh: looks good to me

ozamani9gh (approved):

eve-git (approved):

api/namex/models/request.py
eve-git: Our data shows many duplicate NR submissions within 1-2 minutes. We're making a change to fix duplicates without blocking others. A 5-minute window seems reasonable to us.

ozamani9gh (approved):

PR #1489, opened 2023-11-16
18504 correct the typo of ULCB

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/18504

*Description of changes:*
The typo of "ULCR" instead of "ULCB" causing error in the alteration process (from ULC to BC through NR) and no legalType retrieved .


api/namex/VERSION.py
```diff
@@ -1 +1 @@
-__version__ = '1.1.38'
+__version__ = '1.1.40'
```
eve-git, line 1: Updated the version to 1.1.40 since https://github.com/bcgov/namex/pull/1490 updated the version to 1.1.39.

ozamani9gh (approved):
loos good to me

PR #1497, opened 2023-11-23
18524 After changing the expiration date in name examination, the date is saved incorrectly in Oracle

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/18524

*Description of changes:*
Added the time portion of the expiration date when saving it in oracle.


ozamani9gh (approved):
looks good

PR #1498, opened 2023-11-24
18226 update the condtions of showing reapply button

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/18226

*Description of changes:*
The current conditions for displaying the reapply button are as follows:

The NR state is in ('CONDITIONAL', 'APPROVED').
The expiry date is equal or less than 14 days from now.
The expiry date is after the current date.
Typically, when the expiry date is reached (condition [#3](https://github.com/bcgov/entity/issues/3) not satisfy), a batch job updates the NR state to 'Expired,' making condition [#1](https://github.com/bcgov/entity/issues/1) unsatisfied and preventing the reapply button from showing up.

The expiration process involves setting the NR state to 'Expired' and removing the name from Solr conflict names. However, if the state hasn't been updated to 'Expired' by the batch job upon reaching the expiry date, the names remain in the conflict name list, affecting names cannot be used in other requests. In this case, I believe it is acceptable to allow users to reapply. Therefore, it may be unnecessary to include condition [#3](https://github.com/bcgov/entity/issues/3) above.


ozamani9gh (approved):
looks simple enough

PR #1499, opened 2023-11-27
18687-Updated Namex legaltype

Rajandeep98:
*Issue #18687  NR Type vs Legatype update
https://github.com/bcgov/entity/issues/18687


ozamani9gh (approved):
looks good to me

api/namex/services/lookup/name_request_filing_actions.py
seeker25, line 1: Lines 57-61 needed to change I think

PR #1501, opened 2023-11-28
18567 update nr_day_job query

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/18567

*Description of changes:*
update queries:
1. update the query to pick up the nrs to be expired:
    a. cast expiration_date in the query to be a Date type, so that the time part aligns up CURRENT_DATE, which is 0:00
    b. remove the condition: Request.notifiedBeforeExpiry == True. notified before expiry should not be a condition to expiry an NR.
2. update the query to pick up the nrs to be notified before expired
   a. cast expiration_date in the query to a be Date type 

3. move notify_nr_expired() before notify_nr_before_expiry, because notify_nr_expired includes state_cd update.

### DIFF
--- a/jobs/nr-day-job/nr_day_job.py
+++ b/jobs/nr-day-job/nr_day_job.py
@@ -94,7 +94,8 @@ def notify_nr_before_expiry():
         current_app.logger.debug('entering notify_nr_before_expiry')
 
         where_clause = text(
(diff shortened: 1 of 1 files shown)


ozamani9gh (commented):
looks good to me

ozamani9gh (approved):

PR #1518, opened 2024-04-09
20714 Fix consent email getting sent when conditionally approved NR is reset

semmatti:
*Issue #, if available:*
bcgov/entity#20714

*Description of changes:*
This PR adds a check to only send a consent email if the consent flag has been changed to 'Received'


eve-git (approved):
LGTM

PR #1520, opened 2024-04-19
Partner doesn't response in json format but in xml format, must conve…

stevenc987:
…rt from XML to JSON

*Issue #, if available:* 20909

*Description of changes:*
federal partners have designed MRAS to work with xml – and xml only. We need to covert xml response to json format response.


eve-git (approved):
LGTM

api/namex/resources/mras.py
```diff
@@ -109,6 +109,7 @@ def get(self, province, corp_num):
                 headers=headers
             )
 
+
```
bolyachevets, line 112: 'Accept': 'application/json' header does not work with MRAS API, as it is ignored (unimplemented by the provider of service), so probably needs to switch to xml, just in case provider does not decide to implement json

PR #1533, opened 2024-05-22
removed date formation for expiration date

flutistar:
*Issue #: 20887
*Description of changes:*
Removed unnecessary date formation on `PUT` API.

NOTE: `datetime.datetime.strptime` doesn't work with timezone(%z works properly but %Z doesn't), I converted the date format to `ISO` instead of `UTC`.

Another PR that is made on [Name Examination](https://github.com/bcgov/name-examination/pull/1458) that pairs with this PR.


eve-git (approved):
LGTM

PR #1540, opened 2024-05-31
Added name state type filter to SOLR Query

flutistar:
*Issue #, 20344*

*Description of changes:*
Based on the requirements from name examiners, I added name state filter to display only `A`(Approved) and `R`(Rejected) names on history tab.


EPortman (approved):
Looks good to me

eve-git:
In ticket 20344, the issue was caused by a bracket in the name. How is the code fix related to the name state?

PR #1545, opened 2024-06-04
Fixed error on SOLR Query

flutistar:
*Issue 20344, if available:*

*Description of changes:*
Got SOLR Query from SOLR UI and replaced with wrong one.

### DIFF
--- a/api/namex/analytics/solr.py
+++ b/api/namex/analytics/solr.py
@@ -106,7 +106,7 @@ class SolrQueries:
         HISTORY:
             '/solr/names/select?sow=false&df=name_exact_match&wt=json&&rows={rows}&q={name}'
             '&fl=nr_num,name,score,submit_count,name_state_type_cd,start_date,jurisdiction'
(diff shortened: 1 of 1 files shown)


EPortman:
What is the point of this?

SOLR queries are working in DEV? Just not in local. Why would changing DEV have an effect in local?

flutistar:
> What is the point of this?
> 
> SOLR queries are working in DEV? Just not in local. Why would changing DEV have an effect in local?

I am sure SOLR queries are working in DEV.
I had a pairing session with Richard and SOLR query(with public URL) is working properly.

The only issue is we have wrong ENV variable in local.
@EPortman

EPortman:
If the only issue is a wrong ENV variable in local. What is the point / effect of changing

'&fq=name_state_type_cd:A&fq=name_state_type_cd:R', 
TO
 '&fq=name_state_type_cd:(A%20OR%20R)',

EPortman:
Like in DEV my SOLR is currently working
image

EPortman (approved):

PR #1554, opened 2024-07-04
21835 generate 5 minute delay for email notifications

EPortman:
*Issue #, if available:*
https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/21835

*Description of changes:*
Implemented a debouncer for the email queue to use. 
- This ensures there is a 5 minute delay between when the action occurs and when the email is sent. 
- It also ensures that only the latest email is sent and anytime an action that causes an email is made the 5 minute timer is reset 


EPortman:
For this to work the email address in questions needs to be on the safe list to receive emails in dev / test environments.

EPortman:
Clients are emailed about their NRs in the following scenarios:
1. When NR goes to APPROVE, CONDITIONAL or REJECTED
2. When a payment update is made to REAPPLY or UPGRADE
3. When a NR is refunded
4. When a NR updates to CONSENT_RECEIVED

I added a 5 minute delay for all of these processes. Furthermore if two or more of these occur within 5 minutes of each other it refreshes the timer and only sends the latest email.

eve-git:
My understanding is that the email for a specific NR has a 5-minute delay before being sent. However, upon reviewing your code, it appears that all emails must wait until there has been no activity for 5 minutes, regardless of whether the emails are for different NRs.
Correct me if I am wrong.

EPortman:
You are correct @eve-git . I overlooked this ticket. It is missing the logic preventing users from interrupting each others countdowns  The debouncer needs to be user specific.

Nice catch! Closing this for now.

PR #1555, opened 2024-07-09
generate new nr number

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/21856

*Description of changes:*
Generate a new NR. The logic is the same as in Oracle package NRO_UTIL_PCG.get_new_nr_num


EPortman (commented):
I think there might be a typo here.

```
if random_num < floor or random > ceiling:
         continue
```
         
 Should this be random_num instead of random? Also is this check necessary as the random.randInt gets a random number already in this range?

EPortman (approved):
Looks great!

PR #1559, opened 2024-08-13
remove colin-api

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/19608

*Description of changes:*
1. remove colin-api
2. update the API Resource in colin.py to query from colin-api in lear instead. The endpoint is called from both namerequest and name-examination to validate corp number when changing name.


EPortman (approved):
Looks great.

PR #1564, opened 2024-08-22
fix: ensure expiration date is properly formatted

EPortman:
*Issue #, if available:* [Zenhub 22243](https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/22243)

*Description of changes:*
Expiry dates with milliseconds started appearing in NameX on June 20th, the day after this [release](https://github.com/bcgov/namex/releases) in NameX that included this [ticket related to expiry dates](https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/20887).

On the frontend, the expiration date was converted to ISO format before sending to NameX. on PUT requests
 https://github.com/bcgov/name-examination/pull/1458/files

On the backend, the formatting of expiration date was removed for PUT requests. This was changing UTC format into the correct format for the database - it also truncated milliseconds. 
https://github.com/bcgov/namex/pull/1533/files

Since the format was changed to ISO format on the front end which was adding milliseconds to the expiration date, I added a check if the date is in ISO and to deal with it accordingly. It also works for time in UTC format - either way the date that gets stored will not have milliseconds and be set at the end of the day so that expiration dates remain in the format `2024-10-29 23:59:59+00` when PUT requests occur.


api/namex/resources/requests.py
```diff
@@ -749,6 +749,17 @@ def put(nr, *args, **kwargs):
...
+                expiration_str = json_input['expirationDate']
+                if expiration_str.endswith('UTC'):
+                    parsed_date = datetime.strptime(expiration_str[5:], '%d %b %Y %H:%M:%S %Z')
+                else:
+                    parsed_date = datetime.fromisoformat(expiration_str.rstrip('Z'))
+
+                end_of_day = parsed_date.replace(hour=23, minute=59, second=59)
+                json_input['expirationDate'] = end_of_day.strftime('%Y-%m-%d %H:%M:%S')
+
             # convert NWPTA dates to correct format
```
eve-git, line 763: I recall that the NR expiration date should be set to the end of the day in Pacific Time. Please verify to confirm

EPortman: You are right @eve-git, I missed this part. I just updated it so that when expiration dates are updated they are set to the end of day pacific time. I added checks so that it would work with both UTC and ISO formats.

I tested this locally and now when I update the expiration date the time stays as `YYYY-MM-DD 06:59:00+00` representing the end of day pacific time in UTC format. I went off matching the format to how the expiration date is initially created in `create_expiry_date` https://github.com/bcgov/namex/blob/main/api/namex/services/name_request/abstract_name_request.py

This is keeping expiry dates consistent for me locally. Another option to fix this would be to revert the two PRs I mentioned above so the date stays in UTC format and does not get converted to ISO in the front end. That would require 2 PRs though but would prevent having to handle two different time formats.

eve-git (approved):
LGTM

PR #1610, opened 2024-11-18
#24319 Add business info in the business query from COLIN-API in Entities

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/24319

*Description of changes:*
Add parties and offices info in the business query from COLIN-API in Entities


EPortman (approved):
Looks great

PR #1629, opened 2024-12-11
24849 search endpoint directly queries names db

EPortman:
*Issue #, if available:* https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/24849

*Description of changes:*
When NRO Decommissioning went out **NR numbers** (not the name, the actual number) stopped getting put into SOLR **names core**. Even though firms are not put in the possible conflicts core they were being put in the names core. This caused an inability to find firms on this endpoint - since they are no longer in the possible conflicts core (expected) AND not in the names core (new). 

To get around this the endpoint does not check SOLR anymore, instead it directly queries the names DB with the provided NR number (passed in as query in the query string parameter). 

It seems like this endpoint is only used in this specific use case (couldn't find a call in either NameRequest UI or Name-Examination UI). However testing will needed to be done to confirm this.


api/namex/resources/requests.py
```diff
@@ -388,57 +388,28 @@ def get():
...
-                RequestDAO.stateCd.in_([State.DRAFT, State.INPROGRESS, State.REFUND_REQUESTED]),
-                text(f'({condition})')
+            # Find the NR only if it is not cancelled & not expired or if it has expired within a 60 day grace period. 
+            query = RequestDAO.query.filter(
+                RequestDAO.nrNum == nr_num,
+                RequestDAO.stateCd != State.CANCELLED,
+                or_(
+                    RequestDAO.stateCd != State.EXPIRED,
+                    text(f"(requests.state_cd = '{State.EXPIRED}' AND CAST(requests.expiration_date AS DATE) + "
+                            "interval '60 day' >= CAST(now() AS DATE))")
```
EPortman, line 401: This logic is now directly in the method. It used to call it here:
https://github.com/bcgov/namex/blob/main/api/namex/resources/requests.py#L458

rarmitag (approved):

PR #1631, opened 2024-12-11
24849 search works with name as well

EPortman:
*Issue #, if available:*
https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/24849

*Description of changes:*
In my previous PR, https://github.com/bcgov/namex/pull/1630, I left out the ability to search by name. This brings that functionality back so it works as it should work.
  - Can search (not exact match) by NR number
  - Can search (not exact match) by names in the NR


EPortman:
Hi @severinbeauvais 
Would you mind reviewing this PR when you have a moment? I wanted to provide some context and highlight the changes, as this is the third PR for this issue (apologies for the back-and-forth!).

**Initial Issue with the Endpoint**
  1. /search endpoint would receive a nr_num or name and a rows count
  2. Endpoint initially searches for the NR in Solr using get_parsed_query_name_nr_search
       - Works for most cases but not for firms (not in Solr)
  3. With the results from Solr, it would query the Names DB with added filters for draft, InProgress, or refund
       - Misses firms
  4. While the number of results is less than the rows count and their are results
       - Will search Solr again but this time the results are filtered to not cancelled/expired or within a 60 day expiration grace period
       - Misses firms

**Root Cause**
There are 3 cores in Solr, 2 of which are important here.
  1. The names core: NR number, names and some other stuff.
  2. The possible conflicts core: Same as names core but only names that cause conflicts.

Typically every NR goes into both cores, except for firms that only go into the names core.
NRO decommissioning took this functionality out, now NRs for firms are not in either core - they no longer exist in Solr.
This causes this method to fail.

**Ideas to fix**
  1. Put firms back in the names core
      - This was in a legacy job that has been turned off.
  2. Just search the source of truth (requests table) directly.
      - This was the option these PRs do.

**Changes Made**
PR 1: https://github.com/bcgov/namex/pull/1629/
  - Took out everything Solr, it now searches the requests table directly using the NR num  
 
PR 2: https://github.com/bcgov/namex/pull/1631/
  - Brought back the search by querying the requests table with a LIKE and regex.  
 
PR: 3: This one
  - Brought back functionality to also be able to query the requests table by name.



Concerns

I get the obvious concern is where else this endpoint is getting called from. I cannot find a call to this method at all in any of the names UIs, Services, or Jobs. It seems like the only place this endpoint is used is in this scenario. This needs more testing to confirm. With that being said I cant see a reason why it would cause something else to fail since it is still just sending back matching NRs for a given nr_num or name.

Also, another thing that I took out is that in the original method it prioritized NRs that are in draft, InProgress, or refund state. The reason I took it out was because it added complexity and I could not see a reason why it was necessary. Please correct me if I am wrong and I can add another filter.

Another thing of note that is a more of a names thing is what else uses the names core. Since this core is no longer a complete set of all NRs maybe this is not the only side effect. Have talked to Omid about this and an investigation is under way as to what uses this core. The more important one by far is the possible conflicts core as that is what is queried against when both requesting names and examining names.

severinbeauvais:
Thanks for the great explanation!

Unfortunately, I am not a SME on how Namex (or Business Search/SOLR) work or are used.

I follow what you're saying and share your concerns, especially "how this endpoint is used". I have opinions but I'm not the expert here. Maybe some of the following people have a bigger view of how this is used and what impacts your changes would have:
- @thorwolpert 
- @kialj876 
- @seeker25 
- @vysakh-menon-aot

severinbeauvais:
> Also, another thing that I took out is that in the original method it prioritized NRs that are in draft, InProgress, or refund state. The reason I took it out was because it added complexity and I could not see a reason why it was necessary. Please correct me if I am wrong and I can add another filter.

Prioritizing "recent" NRs makes sense to me when a partial search/match is done, since we care about those more than past NRs.

> Another thing of note that is a more of a names thing is what else uses the names core. Since this core is no longer a complete set of all NRs maybe this is not the only side effect. Have talked to Omid about this and an investigation is under way as to what uses this core. The more important one by far is the possible conflicts core as that is what is queried against when both requesting names and examining names.

1. Is there any reason that firm NRs can be added to SOLR just like the other NRs?
2. Why are NRs in SOLR in the first place? Who uses SOLR (Business Search) to look up NRs? Just the affiliation component in Auth Web?

EPortman:
Thanks for the feedback, @severinbeauvais. 

I spoke with @rarmitag about this and we decided it is probably the better option to revert what I have done and roll it back to what it was before. 

This means every new firm needs to be put into the Names Core. I will start looking into this so that the logic works the way it was before NRO decommissioning. Once this is working I will revert this endpoint back to what it was.

But for the time being:
  -> Dev environment: Can find all NRs only by NR number (not working for name search)
  -> Test environment: Works for everything except firms
  -> Prod environment: Works for everything (NRO decom is not in prod)

severinbeauvais:
Ethan, do you want to merge this PR anyway while awaiting your final solution?

EPortman:
Sure. I guess it cant hurt as its going to be reverted anyway. 

This will allow dev environment to work as expected (can search by Nr num and name, order by most recent, and early return if no query is provided). 

This issue is now a high priority for Names as it has to be done before NRO decommissioning goes up to prod. In a way this fix needs to happen before the end of next week.

severinbeauvais (approved):

PR #1650, opened 2024-12-31
Speed up build

bolyachevets:
*Issue #, if available:*

*Description of changes:* speed up namex-api and emailer builds --

Each line (aka command) in a Dockerfile is a layer. Docker uses a caching mechanism to store intermediate layers. Docker will reuse it from the cache rather than recreating it. This is where the order of instructions in the Dockerfile becomes important. Instructions that change frequently (e.g., copying application code) should be placed near the bottom of the Dockerfile. This ensures that changes in the application code trigger a rebuild of fewer layers, which is faster.


EPortman (approved):
looks good

PR #1656, opened 2025-01-07
update flask-jwt-oidc package

EPortman:
*Issue #, if available:*

*Description of changes:*
The current `flask-jwt-oidc` package was giving the following error when building:
```
RuntimeError: The Poetry configuration is invalid:
  - project must contain ['name'] properties
```

This PR updates this package which changes the `pyproject.toml` file to have these changes:
https://github.com/bolyachevets/flask-jwt-oidc/compare/bump-flask-version...seeker25:flask-jwt-oidc:main


api/pyproject.toml
eve-git, line 1: any reason using the code in seekers fork? Can it change to https://github.com/bcgov/flask-jwt-oidc?

EPortman: It did seem strange to me as well. 

I had originally asked Patrick Wei about this issue and why it started happening with our now without any changes to the package. He was the one that suggested to use this one since it it the one being used in Auth-API:

eve-git (approved):
LGTM

api/pyproject.toml
seeker25: as someone else owns `github.com/bcgov/flask-jwt-oidc`

EPortman: Hi Travis. You are right I should have only updated the lock file for the one package in question instead of running poetry lock generically. 

Is it worth changing the lock file back to what it was and only updating the dependency for the flask-jwt-oidc package?
This build was successful but had some errors in it that I have never seen before:
https://github.com/bcgov/namex/actions/runs/12675035478/job/35325122723#step:4:2089

seeker25: `OP` = 1 password CLI, unrelated I think

EPortman: It seems to be working - will continue to look into it.

Also

I am trying to build the namex-emailer as that also broke recently but am getting the same error with the old flask-jwt-oidc package. What I am confused about is this package isn't referenced anywhere in the requirements file:
https://github.com/bcgov/namex/blob/main/services/emailer/pyproject.toml

But the reason it is giving an error is because it does exist in the poetry lock file:
https://github.com/bcgov/namex/blob/main/services/emailer/poetry.lock#L480

I am assuming it is safe to just remove this reference in the poetry lock since jwt are not used in the NameX Emailer?

This is where the build is failing in the Emailer currently:
https://github.com/bcgov/namex/actions/runs/12675035562/job/35325136264#step:6:1334

PR #1659, opened 2025-01-09
25167 add feature flags service enable won emails

kzdev420:
*Issue #, if available: #25167*

*Description of changes:*

Add Feature Flags service.
Use the enable way of navigation emails FF


api/.env.sample
```diff
@@ -81,3 +81,4 @@ MRAS_SVC_API_KEY=
 
 # Local development only
 DISABLE_NAMEREQUEST_SOLR_UPDATES=1
+LD_SDK_KEY=
```
severinbeauvais, line 84: You can fill this in with the actual Dev key because there's no risk in making it public.

api/flags.json
```diff
@@ -0,0 +1,8 @@
+{
+  "flagValues": {
+      "string-flag": "a string value",
+      "bool-flag": true,
+      "integer-flag": 10,
+      "enable-won-emails": false
```
severinbeauvais, line 6: Don't you only need the last flag, not the sample ones?

kzdev420: For the unit_test, used the sample flags

api/namex/services/flags.py
```diff
@@ -0,0 +1,145 @@
...
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Initialize the Feature Flag environment."""
+        self.app = app
+        self.sdk_key = app.config.get('LD_SDK_KEY')
+
+        if self.sdk_key or app.env != 'production':
+
+            if app.env == 'production':
```
severinbeauvais, line 71: I'm guessing you copied this file from somewhere else, but how can this line be True given line 69?

If this is an exact copy of another file (that works correctly) then I'm OK to leave this instead of introducing risk.

api/flags.json
severinbeauvais: Oh, are the other flags for unit testing?

api/namex/services/flags.py
kzdev420: It's from Legal-Api

services/emailer/devops/vaults.gcp.env
```diff
@@ -22,4 +22,6 @@ BUSINESS_REGISTRY_URL="op://web-url/$APP_ENV/business-registry-ui/BUSINESS_REGIS
 REPORT_API_URL="op://API/$APP_ENV/report-api/REPORT_API_URL"
 REPORT_API_VERSION="op://API/$APP_ENV/report-api/REPORT_API_VERSION"
 LEGAL_API_URL="op://API/$APP_ENV/legal-api/LEGAL_API_URL"
+LEGAL_API_VERSION="op://API/$APP_ENV/legal-api/LEGAL_API_VERSION"
+LD_SDK_KEY="op://launchdarkly/$APP_ENV/namex/LD_SDK_KEY"
```
severinbeauvais, line 26: The LD SDK KEY is actually named differently:

image

api/.env.sample
severinbeauvais: Or leave it blank like the other keys in this file. I don't care :)

services/emailer/config.py
```diff
@@ -90,8 +90,8 @@ class Config:  # pylint: disable=too-few-public-methods
     NOTIFY_API_VERSION = os.getenv("NOTIFY_API_VERSION", "")
     NAMEX_API_URL = os.getenv("NAMEX_API_URL", "")
     NAMEX_API_VERSION = os.getenv("NAMEX_API_VERSION", "")
-    LEGAL_API_URL = os.getenv("NOTIFY_API_URL", "https://legal-api-dev.apps.silver.devops.gov.bc.ca")
-    LEGAL_API_VERSION = os.getenv("NOTIFY_API_VERSION", "/api/v2")
+    LEGAL_API_URL = os.getenv("LEGAL_API_URL", "https://legal-api-dev.apps.silver.devops.gov.bc.ca")
+    LEGAL_API_VERSION = os.getenv("LEGAL_API_VERSION", "/api/v2")
```
severinbeauvais, line 94: This key should be `LEGAL_API_VERSION_2` here and in the example file (and maybe elsewhere).

image

services/emailer/flags.json
severinbeauvais, line 1: Do we need both this file and api/flags.json ?

EPortman:
Once this is ready to be merged the emailer needs to install the latest API changes. 

If simply running `poetry update namex` in the emailer repo is not working I would recommend doing this:
1. Find the commit hash you want the emailer to use
- Run `git log` it will be the first one if you want the latest changes (in this PR)
- Grab the commit hash that it gives (e.g. `4d7464e1a331f53736a5823f71dcde10a5ef2d07`)

2. Update the lock file with this commit hash
- Just update the `resolved_reference` on line 1534
- Example: https://github.com/bcgov/namex/pull/1658/files

services/emailer/src/namex_emailer/services/helpers.py
```diff
@@ -72,39 +74,25 @@ def get_magic_link(nr_number, email, phone):
...
-
-# This function will be removed if the emailer service and NameX API are in sync well 
 @staticmethod
 def get_instruction_group(legal_type, request_action, corpNum):
+    from namex.services import flags  # pylint: disable=import-outside-toplevel
+    enable_won_emails = flags.value('enable-won-emails')
+    structured_log(request, "DEBUG", f"enable way of navigation emails feature flags value: {enable_won_emails}")
+    if enable_won_emails:
+       structured_log(request, "DEBUG", f"enable way of navigation emails")
+       return ReportResource._get_instruction_group(legal_type, request_action, corpNum)
```
severinbeauvais, line 84: How does this implement the new emails when the FF is false?

kzdev420: If FF is False, it will send the old emails

services/emailer/src/namex_emailer/services/helpers.py
```diff
@@ -72,39 +74,25 @@ def get_magic_link(nr_number, email, phone):
...
-    if not corpNum:
-        return False
-    entity_url = f'{current_app.config.get("ENTITY_SVC_URL")}/businesses/{corpNum}'
-    token = get_bearer_token()
-    response = requests.get(entity_url, headers={
-        'Content-Type': 'application/json',
-        'Authorization': 'Bearer ' + token
-    })
-
-    return response
```
severinbeauvais, line 86: Don't you still need  this for CHG/CNV NR emails?

Should you retain the comment below, which seems to apply to the next method?

severinbeauvais (commented):
The purpose of the Feature Flag is to:
- if True, send out the new emails
- if False, send out the old emails

This needs to apply for all "way of navigation" emails -- NEW and CHG/CNV.

I think I don't see both the old and new code, conditional on the FF, in this PR.

services/emailer/src/namex_emailer/services/helpers.py
kzdev420: I think it is the way_of_navigation.
So, it(NEW and CHG/CNV) will be processed when FF is true.

https://github.com/bcgov/namex/blob/4d7464e1a331f53736a5823f71dcde10a5ef2d07/api/namex/resources/name_requests/report_resource.py#L294-L297

vysakh-menon-aot (approved):

services/emailer/src/namex_emailer/services/helpers.py
severinbeauvais: What you told me is this is duplicate code and you're instead using code right in Namex API.

Resolving.

services/emailer/src/namex_emailer/services/helpers.py
severinbeauvais: OK, thanks for the walk-through. Resolving.

api/namex/services/flags.py
severinbeauvais: OK. Resolving. Don't change it.

services/emailer/flags.json
severinbeauvais: Answer: this is used for the emailer. The API flags file is for the API and may not be needed.

Kevin, please ask Ethan whether the API file is really needed.

services/emailer/devops/vaults.gcp.env
severinbeauvais: Changes look good. Resolving.

services/emailer/config.py
severinbeauvais: OK, you've reverted this to v1, so the original key name is ok. Resolving.

services/emailer/flags.json
severinbeauvais: Kevin, double check this before merging.

severinbeauvais:
@kzdev420 Are there new linting and testing issues?

image

severinbeauvais (approved):
LGTM but please double-check per my comments above.

services/emailer/flags.json
EPortman: I don't really know to be honest. I was originally under the impression the flag was going to be used by the API when sending cloud events to the emailer. But from what I can see it is only the emailer that makes use of the flag and the API is left naïve. 

Because of that I guess all of this flag code could of only been in the Emailer and none of it in the API (including this api/flags.json file).

With that being said, I think our team has decided that we want to make use of feature flags going forward after our countless troubles with NRO decom release. And am thinking that having the flag code there is a good thing. And having an example of this file (flags.json) in the api, even if it is redundant, wouldn't be the worst thing.

severinbeauvais: I totallly agree with you, Ethan. Resolving.

And Kevin, thanks again for bring FFs into Namex!

severinbeauvais:
OK, @EPortman and @kzdev420 , I'm leaving this with you to merge and deploy to Dev.

Then, please test this in Dev. Currently, the FF is True. Tomorrow, we'll test again with the FF = False.

image

services/emailer/poetry.lock
EPortman, line 1: Awesome, thanks!

EPortman (approved):

rarmitag (approved):

api/.env.sample
```diff
@@ -81,3 +81,6 @@ MRAS_SVC_API_KEY=
 
 # Local development only
 DISABLE_NAMEREQUEST_SOLR_UPDATES=1
+
+# launchdarkly
+NAMEX_LD_SDK_ID=sdk-075200eb-4ff7-4e0e-872a-848585d3d460
```
severinbeauvais, line 86: @kzdev420 
I bet this key has to be imported in api/config.py.

PR #1676, opened 2025-01-16
25270_add_env_values

kzdev420:
*Issue #, if available: #25270*

*Description of changes:*


services/emailer/config.py
```diff
@@ -105,6 +105,10 @@ class Config:  # pylint: disable=too-few-public-methods
     ACCOUNT_SVC_CLIENT_SECRET = os.getenv("KEYCLOAK_CLIENT_SECRET")
     ACCOUNT_SVC_TIMEOUT = os.getenv("KEYCLOAK_TIMEOUT")
 
+    ENTITY_SVC_AUTH_URL = os.getenv("KEYCLOAK_AUTH_TOKEN_URL")
+    ENTITY_SERVICE_ACCOUNT_CLIENT_ID = os.getenv("KEYCLOAK_CLIENT_ID")
+    ENTITY_SERVICE_ACCOUNT_CLIENT_SECRET = os.getenv("KEYCLOAK_CLIENT_SECRET")
```
EPortman, line 110: Since these values are defined already, and thus redundant. 

Do you think you could just change where ENTITY_SVC_AUTH_URL, ENTITY_SERVICE_ACCOUNT_CLIENT_ID, and ENTITY_SERVICE_ACCOUNT_CLIENT_SECRET are referenced to be ACCOUNT_SVC_AUTH_URL, ENTITY_SERVICE_ACCOUNT_CLIENT_ID, and ENTITY_SERVICE_ACCOUNT_CLIENT_SECRET?

kzdev420: I don't think it can be changed because it's in the API and that function can be referenced from somewhere.

https://github.com/bcgov/namex/blob/58238aaf5facc5877a003b1fed4f72a087580f2c/api/namex/resources/utils.py#L33

EPortman (approved):

services/emailer/config.py
EPortman: Okay, its strange because it was working before and the new code changes caused it to break. 

But I merged it either way.

PR #1679, opened 2025-01-17
25356 update_outputs_CHG_CNV_NR

kzdev420:
*Issue #, if available: bcgov/entity#25356*

*Description of changes:*

### DIFF
--- a/services/emailer/poetry.lock
+++ b/services/emailer/poetry.lock
@@ -1531,7 +1531,7 @@ zipp = "^3.8.1"
 type = "git"
 url = "https://github.com/bcgov/namex.git"
 reference = "HEAD"
-resolved_reference = "5e07bca426b4813b36b75fda4adfbe4fa51bdc88"
+resolved_reference = "810650b8fc1e4c1b83d5ba7347eda4176bf5915e"
 subdirectory = "api"
 
 [[package]]
--- a/api/namex/resources/name_requests/report_resource.py
+++ b/api/namex/resources/name_requests/report_resource.py
@@ -173,7 +173,8 @@ class ReportResource(Resource):
         if nr_report_json['applicants']['countryTypeCd']:
             nr_report_json['applicants']['countryName'] = \
                 pycountry.countries.search_fuzzy(nr_report_json['applicants']['countryTypeCd'])[0].name
-        actions_obj = ReportResource._get_next_action_text(nr_model['entity_type_cd'])
+        action_url = ReportResource._get_action_url(nr_model['entity_type_cd'], instruction_group)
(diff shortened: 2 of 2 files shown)


services/emailer/poetry.lock
```diff
@@ -1531,7 +1531,7 @@ zipp = "^3.8.1"
 type = "git"
 url = "https://github.com/bcgov/namex.git"
 reference = "HEAD"
-resolved_reference = "5e07bca426b4813b36b75fda4adfbe4fa51bdc88"
+resolved_reference = "810650b8fc1e4c1b83d5ba7347eda4176bf5915e"
```
severinbeauvais, line 1534: Why does this need to be updated? Aren't you pointing to the same old Namex API?

severinbeauvais (approved):
I believe this is OK but I'd like to see a series of successful tests:
- CHG or CNV NR for BC
- CHG or CNV NR for CR/CC/ULC in LEAR
- CHG or CNV NR for CR/CC/ULC in COLIN
- regression tests (FF on/off, NEW NRs, etc)

services/emailer/poetry.lock
kzdev420: I updated the NameX API code.
So, It should be the new NameX API.

severinbeauvais: This needs to work in Prod with the old API.

kzdev420: Yes, no problem. It will work in Prod with the old API

stevenc987 (approved):

services/emailer/poetry.lock
severinbeauvais: > I updated the NameX API code. So, It should be the new NameX API.

> Yes, no problem. It will work in Prod with the old API

Are these both True?

severinbeauvais: Kevin said:
> it should be updated. because the Emailer installs and uses the API as the dependency from the git repository.

severinbeauvais: @ethantspitt , are you OK with this change?

severinbeauvais (approved):

services/emailer/poetry.lock
EPortman: Yes updating this reference makes sense to me. 

The reason being:
 - The emailer imports the `ReportResource` from NameX as a dependency: https://github.com/bcgov/namex/blob/main/services/emailer/src/namex_emailer/email_processors/nr_notification.py#L27
 - Since this code was updated the emailer needs to import this change so that the emailer can make use of it.
 - If it was an endpoint on the API that changed which the emailer called, however, it would be a totally different story.

PR #1707, opened 2025-02-19
25237 Fixed continued in entity type codes

severinbeauvais:
*Issue #:* bcgov/entity#25937

*Description of changes:*
- added parameter to get_actions()
- updated raw_nr_to_action_mapping with actual entity type code used
- updated get_actions to use request action code to build continued in dict key
- updated unit test
- updated Namex API version to 1.2.28


api/namex/services/lookup/name_request_filing_actions.py
```diff
@@ -129,31 +129,31 @@ def teardown(self, exception):
...
-        ('CCR_C', 'C', 'Continued In BC Limited Company', 'Change of Name', 'lear', None, None, None),
-        ('RCR_C', 'C', 'Continued In BC Limited Company', 'Restoration', 'static', None, 'https://www2.gov.bc.ca/gov/content/employment-business/business/managing-a-business/permits-licences/businesses-incorporated-companies/forms-corporate-registry#information-packages', None),
-        ('BECR_C', 'C', 'Continued In BC Limited Company', 'Conversion from Continued In Benefit Company ', 'lear', 'alteration', None, "{'filing': {'header': {'name': {entitiesFilingName} }, '{entitiesFilingName}': {'nameRequest': {'nrNumber': '{nrNumber}', 'legalName': '{legalName}', 'legalType': '{legalType}'}}}}"),
-        ('ULCB_C', 'C', 'Continued In BC Limited Company', 'Conversion from Continued In ULC', 'lear', 'alteration', None, "{'filing': {'header': {'name': {entitiesFilingName} }, '{entitiesFilingName}': {'nameRequest': {'nrNumber': '{nrNumber}', 'legalName': '{legalName}', 'legalType': '{legalType}'}}}}"),
+        ('CT_CR', 'C', 'Continued In BC Limited Company', 'Continuation In', 'lear', None, None, None),
```
severinbeauvais, line 132: "If this is a NR of type CT, for entity type CR, then map legalType = 'C'."

api/namex/services/lookup/name_request_filing_actions.py
```diff
@@ -187,11 +187,11 @@ def get_actions(self, nr_type: str, entity_type_cd: str) -> Optional[dict]:
...
-        if nr_type in ['FR', 'CFR'] and entity_type_cd in ['FR', 'DBA', 'GP']:
-            nr_type = nr_type + '_' + entity_type_cd
+        if request_type_cd in ['FR', 'CFR'] and entity_type_cd in ['FR', 'DBA', 'GP']:
+            request_type_cd = request_type_cd + '_' + entity_type_cd
 
         # special case for continued in types that otherwise have the same NR type as regular companies
-        if entity_type_cd in ['C', 'CBEN', 'CCC', 'CUL']:
-            nr_type = nr_type + '_' + entity_type_cd
+        if request_action_cd == 'MVE':
+            request_type_cd = request_type_cd + '_' + entity_type_cd
```
severinbeauvais, line 195: "If the action is MVE (cont in) then look up the entry with request type + entity type."
Eg,
request action = MVE
request type = CT
entity type = CR
look up =>"CT_CR"

severinbeauvais:
@eve-git @EPortman @ozamani9 @stevenc987 , please review. Thanks!

EPortman (approved):
looks great to me. Thanks for fixing some of the linting errors too. 

The only thing missing is the version number bump.

eve-git (approved):

api/namex/services/lookup/name_request_filing_actions.py
```diff
@@ -187,11 +187,15 @@ def get_actions(self, nr_type: str, entity_type_cd: str) -> Optional[dict]:
...
 
-        # special case for continued in types that otherwise have the same NR type as regular companies
+        # special case for continued in request action since entity_type_cd is CR/BC/CC/ULC instead of C/CBEN/CCC/CUL
+        if request_action_cd == 'MVE':
+            request_type_cd = request_type_cd + '_' + entity_type_cd
+
+        # special case for continued in entity types that otherwise have the same NR type as regular companies
         if entity_type_cd in ['C', 'CBEN', 'CCC', 'CUL']:
-            nr_type = nr_type + '_' + entity_type_cd
+            request_type_cd = request_type_cd + '_' + entity_type_cd
```
severinbeauvais, line 199: "If the entity type is C (continued in CR) then look up the entity with request type + entity type."
Eg,
request type = CCR
entity type = C
look up => "CCR_C"

api/namex/services/lookup/name_request_filing_actions.py
severinbeauvais: Namerequest UI should provide the correct entity type (eg, C instead of CR) for a MVE NR but I'm afraid that might break COLIN.

On the other hand, I don't think any of the continued in NRs worked before I added the cont in entries in the table above in the last major release.

api/namex/services/lookup/name_request_filing_actions.py
```diff
@@ -187,11 +191,15 @@ def get_actions(self, nr_type: str, entity_type_cd: str) -> Optional[dict]:
...
-        # special case for continued in types that otherwise have the same NR type as regular companies
+        # special case for continued in entity types that otherwise have the same NR type as regular companies
         if entity_type_cd in ['C', 'CBEN', 'CCC', 'CUL']:
-            nr_type = nr_type + '_' + entity_type_cd
+            request_type_cd = request_type_cd + '_' + entity_type_cd
 
-        return self.get_dict.get(nr_type)
+        # special case for continued in request action since entity_type_cd is CR/BC/CC/ULC instead of C/CBEN/CCC/CUL
+        if request_action_cd == 'MVE':
+            request_type_cd = request_type_cd + '_' + entity_type_cd
```
severinbeauvais, line 203: I decided to move this special case after the other one, in case Namerequest UI is changed to provide the expected C/CBEN/CCC/CUL entity type.

If Namerequest UI is fixed then the original entries in the table above get used.

While Namerequest UI is not fixed, the extra entries in the table (and this special case) get used.

api/namex/services/lookup/name_request_filing_actions.py
severinbeauvais: See my other comment regarding Namerequest UI.

I think the code fix could have been in either Namerequest UI or here. This PR fixes it here but will also work if Namerequest UI is changed :)

PR #1715, opened 2025-02-28
513 create ocp sftp relay api

EPortman:
*Issue #, if available:* [zenhub ticket](https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/business-ar/513)

*Description of changes:*
Namex has a [job that uses the sftp protocol](https://github.com/bcgov/namex/tree/main/jobs/sftp-nuans-report). This job is about to be moved from OCP to GCP.

Since GCP does not have a connection to the the BcGov Network, the SFTP protocol will not work. This api takes the code from the job that manages the SFTP connection and wraps it behind an endpoint with authentication. The end result is that the SFTP job will be able to call this endpoint and send a file to be uploaded without requiring a connection to the BcGov Network.

This is how the endpoint is called:
```
curl -X POST \
  -H "Authorization: Bearer <token>" \
  -F "file=<some_file.zip>" \
  <base_url>/api/v1/sftp/upload
```


rarmitag:
Appearently theres a line in auth.py that Sonarcloud wants you to set to true (verify_signature of token)

EPortman:
> Appearently theres a line in auth.py that Sonarcloud wants you to set to true (verify_signature of token)

Yeah noticed that, will update code to get rid of the error. Thanks.

PR #1725, opened 2025-03-03
26099 refactor_email_template_organization

kzdev420:
*Issue #:* bcgov/entity#26099

*Description of changes:*


services/emailer/src/namex_emailer/email_processors/__init__.py
```diff
@@ -46,3 +47,48 @@ def substitute_template_parts(template_code: str) -> str:
...
+    # Check the request_action template first
+    template_path = base_path / request_action / template_name
+    if template_path.exists():
+        return template_path.read_text()
+
+    # Check the specific status-based template
+    if status:
+        template_path = base_path / request_action / status / template_name
+        if template_path.exists():
+            return template_path.read_text()
```
kzdev420, line 79: Get the templates from the refactored directory.

services/emailer/src/namex_emailer/email_processors/__init__.py
```diff
@@ -46,3 +47,48 @@ def substitute_template_parts(template_code: str) -> str:
...
+    # Check the common template fallback
+    common_template_path = base_path / "common" / template_name
+    if common_template_path.exists():
+        return common_template_path.read_text()
+
+    # Check the status-based common template
+    if status:
+        common_template_path = base_path / "common" / status / template_name
+        if common_template_path.exists():
+            return common_template_path.read_text()
```
kzdev420, line 90: Get the old template

services/emailer/src/namex_emailer/email_templates/ASSUMED/NR-EXPIRED.html
severinbeauvais, line 1: From here down, these are copies of the original template, but maybe specific to the NR type, right?

severinbeauvais (commented):
This is explicit but I like it -- easy to look at all different flavours of emails per NR type. The only downside is that several templates might need to be updated due to duplication. I think I'm OK with that.

Ethan, Steven, Eve, what do you think?

Kevin, which of these did you test? One of each? Can you show your test results, please and thanks?

EPortman:
> This is explicit but I like it -- easy to look at all different flavours of emails per NR type. The only downside is that several templates might need to be updated due to duplication. I think I'm OK with that.
> 
> Ethan, Steven, Eve, what do you think?
> 
> Kevin, which of these did you test? One of each? Can you show your test results, please and thanks?


I agree for the most part. Easier to work with and debug.

Although with the volume of file changes / duplication and seemingly non-existent local tests it could be a nightmare to test all of these properly.

severinbeauvais:
> Although with the volume of file changes / duplication and seemingly non-existent local tests it could be a nightmare to test all of these properly.

May I propose a unit test to verify that the correct "main template" is selected for each possible variation?

services/emailer/tests/unit/email_processors/test_get_template.py
```diff
@@ -0,0 +1,57 @@
...
+    if not response:
+        assert result is None
+        mock_log.assert_called_once_with(
+            request, "ERROR", f"Failed to get {request_action}, {status}, {template_name} email template"
+        )
+    else:
+        assert isinstance(result, str)
+        if response == "common":
+            mock_log.assert_called_once_with(
+                request, "DEBUG", f"Not Found the template from {request_action}/{status}/{template_name}"
```
severinbeauvais, line 56: Thanks for the unit test! I'm glad you found a code improvement as a result of it.

What do you think about fully populating the test parameters (ie, test all possible NR types and states)? This would validate that your "333 files changed in this PR" have the right names in the right folders (since it's not practical to review all of them in this PR, haha).

severinbeauvais (approved):

EPortman (approved):
Looks great. 

This can be merged by the way as it is not part of the GCP migration.

severinbeauvais:
@ethantspitt , are you OK to merge this PR now?

EPortman:
I am yes, it all looks great.

PR #1727, opened 2025-03-03
25237 fix a small bug with legal type of CULs

JazzarKarim:
Issue #: https://github.com/bcgov/entity/issues/25937

*Description of changes:*
There was a small bug that was causing issues with the legal type of CULs, from FE:
image

`legaltype` is undefined. This should fix it.


severinbeauvais (approved):
Yes, this should fix the bug.

This will need to be deployed to Test asap, @ozamani9gh .

EPortman (approved):

PR #1753, opened 2025-03-13
26412_add_continuation_in_email

kzdev420:
*Issue #, if available: #26412*

*Description of changes:*


services/emailer/src/namex_emailer/services/helpers.py
```diff
@@ -59,13 +59,41 @@ def query_nr_number(identifier: str):
...
-        'nr': nr_number,
-        'email': email,
-        'phone': phone
+        'nr': nr_data['nrNum'],
+        'email': emails[0],
+        'phone': phones[0]
     }
     encoded_params = urlencode(params)
-    return f'{BUSINESS_REGISTRY_URL}incorporateNow/?{encoded_params}'
+    return f'{BUSINESS_REGISTRY_URL}{route}/?{encoded_params}'
```
kzdev420, line 79: Generate the magic link based on the NR type

api/namex/resources/name_requests/report_resource.py
```diff
@@ -293,32 +293,24 @@ def _is_potential_colin(legal_type):
...
         if request_action in {RequestAction.REH.value, RequestAction.REST.value}:
             return 'reh'
         if request_action in {RequestAction.CHG.value, RequestAction.CNV.value}:
             # For the 'Name Change' or 'Alteration', return 'modernized' if the company is in LEAR, and 'colin' if not
             return 'modernized' if ReportResource._is_lear_entity(corpNum) else 'colin'
-        if not ReportResource._get_enable_won_emails_ff():
-            return ReportResource._old_get_instruction_group(legal_type)
         if ReportResource._is_modernized(legal_type):
+            if request_action == RequestAction.MVE.value and not email_feature_flags.get('enable_cont_in_emails'):
+                return 'modernized-old'
```
kzdev420, line 305: Return `modernized`(new-version) if the `enable_cont_in_emails` FF is True.
Return `modernized-old`(old-version) if the `enable_cont_in_emails` FF is False.

severinbeauvais: There's another FF, `enable-amalgamation-emails` that you could integrate now.

services/emailer/src/namex_emailer/email_templates/template-parts/name-request/nr-footer.html
```diff
@@ -1,6 +1,6 @@
 **Business Registry**
 BC Registries and Online Services
 
-Toll Free: 1-877-370-1033
-Victoria Office: 250-370-1033
+Toll Free: [1-877-370-1033](1-877-370-1033)
+Victoria Office: [250-370-1033](250-370-1033)
```
severinbeauvais, line 5: Can you please show sample output?

services/emailer/src/namex_emailer/services/helpers.py
```diff
@@ -59,13 +59,41 @@ def query_nr_number(identifier: str):
...
 @staticmethod
-def get_magic_link(nr_number, email, phone):
+def get_magic_link(nr_data):
     """Return a magic link."""
     BUSINESS_REGISTRY_URL = current_app.config.get("BUSINESS_REGISTRY_URL")
+    magic_link_route = {
+        'NEW': 'incorporateNow',
+        'MVE': 'continueInNow',
+        'AML': 'amalgamateNow',
+        # 'REN': 'registerNow'
```
severinbeauvais, line 69: `REN` is for "renew".

"registerNow" is for NEW firm NRs (instead of coop or corp).

I'd like to see logic for NEW firm NRs, as we will likely add the magic link to their emails in the future.

severinbeauvais:
Reviewers, can you please look at this today? We'd like it in Dev (and ideally Test) by end of day so it can be verified to see if any other changes are needed for Apr 1. Thanks.

EPortman (approved):
This looks good to me. 

There might need to be some coordination about whenever this goes to prod, are all of these changes only to go in on April 1st or can it go in earlier?

The reason I say this is we are currently experiencing quite a large bug in dev and test where email confirmations / receipts are not going out when creating an NR. They are working now in prod though. 

The last time I got one of these confirmation emails was on February 24th but I don't normally put in my full email address so am unsure as to when they stopped working. Am working on this today but since the emailer calls NameX API it needs to be deployed to prod which it cannot right now because of this issue.

severinbeauvais:
> There might need to be some coordination about whenever this goes to prod, are all of these changes only to go in on April 1st or can it go in earlier?

This should  be able to go in earlier, because the new emails are disabled by FF. The email issues are probably environmental (though we could check Namex Emailer logs if you have a time to search for).

Note that we will need to rework the FF logic for MVP (planned for mid May) as we will no longer tell folks to go to COLIN. This will be future work. cc: @Mihai-QuickSilverDev

EPortman:
Okay, merging this. 

The work for this receipt issue is happening on Monday. I was just wanting to bring it to attention in case anyone has gotten a receipt email more recent than Feb 24th. The logs in the emailer do clearly say `Failed to get payment info for name request id: 2270467` as an example which is this code here https://github.com/bcgov/namex/blob/main/services/emailer/src/namex_emailer/email_processors/name_request.py#L83.

I agree though it probably is an environment issue it is just strange that this is the only email that is not working.

EPortman:
Like, im probably missing something obvious but this is in the Emailer logs for that call
image

And that log should only fire if the response is not okay, but then the matching log in the API says that it did return a 200
image


But this is a Names issue and will be sorted out next Monday.

severinbeauvais:
I just create a continued in NR and a new NR. I received both "results of Name Request" emails. These contain a PDF with the result, but no receipts. So, email is sort of working.

Note that I got a "payment incomplete" error for the first NR, but when I retried payment it all completed by itself (I did not have to re-enter payment info). (There is a bug ticket for this.) Maybe the Emailer tries to send the email while the payment is still incomplete???

image

severinbeauvais:
@riyazuddinsyed , are you able to test this ticket in Dev? Mihai would like it in Test by end of today.

EPortman:
Thank you for checking Severin, I am actually getting the same behavior as you now. I also suspect that the emailer is trying to send the email prematurely. 😬

Will be a top priority next week.

PR #1754, opened 2025-03-13
fix env var typo and prepend namex before job name

EPortman:
*Issue #, if available:*
Values pulled from env need to match what is in the vaults.gcp.env file
https://github.com/bcgov/namex/blob/main/jobs/notebook-report/devops/vaults.gcp.env#L14

*Description of changes:*


eve-git (approved):
LGTM

PR #1755, opened 2025-03-14
Solr Feeder - re-add oracle sql documentation

kialj876:
Issue ref: #26311

*Description of changes:*
- This got removed at some point but it is still relevant / helpful (we had a search bug come up to do with this stuff earlier this week and I had a really hard time remembering what to look for in COLIN without this)
- It used to be in the folder 'nro-legacy' which was misleading and that's probably why it got removed. I've added it inside the solr-feeder folder because it is only relevant to the solr-feeder flow
- NOTE: some of this may be slightly out of date, but it is generally correct and still very helpful for figuring out where to look in CPRD when syncing issues come up from the COLIN side


EPortman (approved):
looks good

PR #1756, opened 2025-03-14
#25261 Explicitly call db.session.commit() when saving edits

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/25261

*Description of changes:*
Explicitly call db.session.commit() when saving edits to ensure changes are not missed.

### DIFF
--- a/api/namex/resources/name_requests/name_request.py
+++ b/api/namex/resources/name_requests/name_request.py
@@ -4,7 +4,7 @@ from uuid import uuid4
 import requests
 from flask import current_app, jsonify, request, make_response
 
-from namex import jwt
+from namex import jwt, db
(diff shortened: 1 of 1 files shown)


api/namex/resources/name_requests/name_request.py
```diff
@@ -320,6 +320,11 @@ def handle_patch_edit(self, nr_model: Request):
 
         # This handles updates if the NR state is 'patchable'
         nr_model = self.update_nr(nr_model, nr_model.stateCd, self.handle_nr_patch)
+        # Commit the transaction to ensure changes are saved
```
eve-git, line 323: It is the second change for #25261, no need to update version number again.

EPortman (approved):

PR #1759, opened 2025-03-14
26412 cont_in_magic_link_email

kzdev420:
*Issue #, if available: #26412*

*Description of changes:*


EPortman (approved):
I am okay with these changes

api/namex/resources/name_requests/report_resource.py
```diff
@@ -278,7 +278,7 @@ def _is_modernized(legal_type):
 
     @staticmethod
     def _is_colin(legal_type):
-        colin_list = ['XCR', 'XUL', 'RLC']
+        colin_list = ['XCR', 'XUL', 'RLC', 'CR', 'UL', 'CC']
```
severinbeauvais, line 281: What NRs / conditions is this list used for?

kzdev420: All NRs can use this list, but except the NR type to be the New or Cont_In and  FF to be true.

`CR, UL, CC` were the potential_colin.
If NR type(`NEW, Cont_in`) or FF is false, it did return 'colin'.
So, I combined the colin and potential colin list, and this if_condition moved lowest level.
And, I created a new list (`CR, UL, and CC`) to check the NRs want the magic link, and magic_link_if_condition(FF and NR types(`New, Cont_In`)) is placed higher than colin_if_condition

severinbeauvais: OK, sounds good.

PR #1779, opened 2025-04-03
26984 namex emailer bug

EPortman:
## Description of Changes

This PR addresses an issue in the `ReportResource._get_instruction_group` method in NameX API, which determines feature flags and the decision tree for the emailer. Specifically, it fixes the `_is_magic_link` method so it correctly handles request actions beyond the two originally supported (`NEW` and `MVE`).

### Code Changes
- **API**  
  - **Bug Fix**: Updated `_is_magic_link` to use a default empty list when the `request_action` key is missing from `magic_link_list`. This ensures the method always returns a boolean instead of raising an error.  
  - **Context**: Previously, `magic_link_list.get(request_action)` returned `None` for unrecognized actions, leading to `legal_type in None`, which caused the `TypeError`, which was getting caught [here](https://github.com/bcgov/namex/blob/main/services/emailer/src/namex_emailer/email_processors/nr_result.py#L131) leading to a response object to be returned and ultimately a error when trying to iterate the response object [here](https://github.com/bcgov/namex/blob/main/services/emailer/src/namex_emailer/resources/worker.py#L74)

- **Emailer**  
  - **Dependency Update**: Updated the emailer to reference the new API commit that includes this bug fix.  
  - **Logging Improvements**: Added additional logging to aid in debugging.

**It is important to NOT Squash and Merge this PR as the first commit hash created here is needed and squashing it will combine the two into a new commit hash.**


services/emailer/poetry.lock
```diff
@@ -1754,7 +1754,7 @@ zipp = "^3.8.1"
 type = "git"
 url = "https://github.com/bcgov/namex.git"
 reference = "HEAD"
-resolved_reference = "6574bf53d46c90fc2def969663f13ca0eacc05fe"
+resolved_reference = "8399615407ad6e0d5601957f61cd2d59c972bbc2"
```
EPortman, line 1757: This is telling the emailer to install the API at this commit: `26984: fix is_magic_link method in api for emailer`

services/emailer/src/namex_emailer/email_processors/nr_result.py
```diff
@@ -128,6 +130,7 @@ def email_report(email_info: SimpleCloudEvent):
         email['content']['attachments'] = attachments
         return email
     except Exception as err:
+        structured_log(request, "ERROR", f"Error retrieving the report: {traceback.format_exc()}")
```
EPortman, line 133: This prints the error stack trace when something goes wrong in the logic flow.

api/namex/resources/name_requests/report_resource.py
```diff
@@ -291,7 +291,7 @@ def _is_magic_link(legal_type, request_action):
             'NEW': ['CR', 'UL', 'CC'],
             'MVE': ['CR', 'UL', 'CC']
         }
-        return legal_type in magic_link_list.get(request_action)
+        return legal_type in magic_link_list.get(request_action, [])
```
EPortman, line 294: Fixes a type error when request_action is neither `NEW` nor `MVE`.

eve-git (approved):
LGTM

severinbeauvais (commented):
I have reviewed this. Good find. Thanks!

PR #1794, opened 2025-04-22
26088 - Added Optional filters params for search

Rajandeep98:
*Issue #, if available:* bcgov/entity/issues/26088
*Description of changes:*
added filters for search


eve-git (approved):
LGTM, please update api version

eve-git (approved):

api/namex/resources/requests.py
```diff
@@ -485,11 +485,23 @@ def _get_next_set_from_solr(solr_query, start, rows):
...
+
+        # Add the state filter if 'state' is provided
+        if state:
+            q = q.filter(RequestDAO.stateCd.in_(state))
+
+        # Add the nr_name filter if 'nr_name' is provided
+        if nr_name:
+            q = q.filter(RequestDAO.nameSearch.ilike(f'%{nr_name}%'))
+            
+        q = q.options(
```
severinbeauvais, line 504: For comparison, what's the response time of the old code (with "options" as part of the query)?

I see in your screenshot that the new call took 704 ms, which seems kind of slow...?

severinbeauvais (commented):
LGTM. Please test thoroughly so this doesn't break the existing feature.

Also, are you leaving it to Auth API to figure out the pagination, or can those parameters be passed into here?

We're going to have to think about how to combine NR results and business results. I'm thinking, if page size is, say 100, that this call would return up to 100 results and the Legal API call would return up to 100 results. Thoughts?

JazzarKarim (approved):
Looks good.

Rajandeep98:
for now, pagination can be implemented on auth side

severinbeauvais:
> for now, pagination can be implemented on auth side

I think pagination on this side would result in a large efficiency/speed improvement, especially when no filters are used. What do you think? Is there a design/plan for combining NRs and businesses into a single page?

EPortman (approved):

PR #1801, opened 2025-04-29
20384 build resend notification

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/20384

*Description of changes:*
1. log notifications in events table when first sent
2. retrieve the notification events and resend

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the namex license (Apache 2.0).

PR #1802, opened 2025-04-29
27414 - Fix Before-Expiry NR Email Logging Error

meawong:
Issue #: /https://github.com/bcgov/entity/issues/27414 

*Description of changes:*
- The first call to `get_main_template` was causing logging errors since it doesn't have the correct values for the **before-expiry NR case** => moved this call to the else case

Logs Before:
image

Logs After:
image

### DIFF
--- a/services/emailer/src/namex_emailer/email_processors/nr_notification.py
+++ b/services/emailer/src/namex_emailer/email_processors/nr_notification.py
@@ -85,7 +85,6 @@ def process(email_info: SimpleCloudEvent, option) -> dict:  # pylint: disable-ms
 
     file_name_suffix = option.upper()
     request_action = nr_data["request_action_cd"]
-    template = get_main_template(request_action, f"NR-{file_name_suffix}.html")
     if option == Option.BEFORE_EXPIRY.value:
         if "entity_type_cd" in nr_data:
             legal_type = nr_data["entity_type_cd"]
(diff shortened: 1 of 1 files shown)


EPortman (approved):
Looks great! Exactly what I had in mind. 

Thank you for taking the time to run the emailer locally 🙏🙏

EPortman:
Hmm, I didn't even think about that. Good point. 

@eve-git do you mind holding off merging https://github.com/bcgov/namex/pull/1801/ until this has made it to test?

And then maybe just release this and only this and create a tiny release-report for it? I don't know what is best practice to do here.. Especially since nothing is actually broken (as far as the business is concerned), so don't know if this warrants a hot-fix release.

severinbeauvais:
> And then maybe just release this and only this and create a tiny release-report for it? I don't know what is best practice to do here.. Especially since nothing is actually broken (as far as the business is concerned), so don't know if this warrants a hot-fix release.

Entity Team's "hotfix process" is to push single fixes through without the release overhead. I don't mind that here... do we need any special testing to verify that this fix didn't break something else? @meawong

eve-git:
@EPortman Sorry, I merged it before I saw your message. However, my commit will only be deployed to DEV and is not ready to be deployed to TEST yet.

meawong:
> > And then maybe just release this and only this and create a tiny release-report for it? I don't know what is best practice to do here.. Especially since nothing is actually broken (as far as the business is concerned), so don't know if this warrants a hot-fix release.
> 
> Entity Team's "hotfix process" is to push single fixes through without the release overhead. I don't mind that here... do we need any special testing to verify that this fix didn't break something else? @meawong

I did some spot testing with different NR types which looked good but I don't see any special testing required. @severinbeauvais

EPortman:
No problem @eve-git. 

But now the emailer in dev holds both this commit and 1801. And sending one to test means sending the other. 

Since this is not a business issue (before-expiry emails are still getting sent), we might just have to live with the logging errors in the emailer for now until the next proper release. Which I am fine with. I don't think anything bad will come of it except maybe Ali might get notified on pager duty since they are structured logging errors.

PR #1804, opened 2025-05-01
20384r esend notification code update

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/20384

*Description of changes:*
I wanted to minimize testing effort for now, so I’ve kept the original code largely untouched while enabling the ability to resend notifications.

Changes made:

1. Code refactor – Moved the Option Enum to a constants file.

2. Logic updates:

    a. When a notification is sent, the email content is saved to the events table. Instead of saving the entire attachment report as byte in event_json, I save nr_model, allowing us to avoid storing large files and regenerate the report later as needed.

    b. The report is re-generated dynamically from the nr_model.

    c. Notification can be resent by retrieving directly from the event and the attachment is regenerated based on the nr_model.


EPortman (approved):
lgtm

PR #1807, opened 2025-05-13
Updates LEGAL_API_VERSION to LEGAL_API_VERSION_2.

ahamood1:
*Issue 23558:*

- Please refer to the steps to address [the warning as deprecated](https://github.com/bcgov/entity/issues/23558#issuecomment-2825428887)

*Description of changes:*

- Changes env var to use /api/v2.
- LEGAL_ API_VERSION -> LEGAL_API_VERSION_2


EPortman (approved):
good stuff. 

BTW versions are updated manually in Namex. So you would want to bump this to `1.2.42`
https://github.com/bcgov/namex/blob/main/api/namex/VERSION.py#L1

Its easily forgotten though.

eve-git (changes requested):
I believe the purpose of the environment variable is to provide flexibility. At this point, renaming it from LEGAL_API_VERSION to LEGAL_API_VERSION_2 doesn’t make much sense—we just need to update its value.

ahamood1:
@eve-git, as stated before and in the PR, in 1password their was already a variable that has the new version of the api end point '/api/v2' and it was saved under LEGAL_API_VERSION_2. I am not sure why you think I just renamed the variable and kept the same value...

eve-git:
Thanks for the update, Ali. I don’t have access to 1Password. If the variable name is shared across multiple apps, I agree—we should update it on our side as well.

eve-git (approved):

eve-git:
By the way, Ali, could you talk to Severin when they are going to move to V2 entirely? I still feel not good update the variable name -- we don't get any waring so far, do we really need to move to V2?

ahamood1:
@eve-git please see image below.

Screenshot 2025-05-14 at 9 16 30 AM

ahamood1:
Maybe I misunderstood ticket [23558](https://github.com/bcgov/entity/issues/23558) please clarify.

eve-git:
I knew I had commented in the ticket, but I wasn’t aware that there are two variables in 1Password—one for V1 and another for V2. That suggests some apps may still be using V1. In that case, we likely don’t need to update the version on our end, especially since no warnings have occurred in Namex so far.

Once they fully abandon V1, they can update the shared variable’s value to V2. At that point, we’ll just need to test the change—no code updates should be necessary.

severinbeauvais (approved):
LGTM

api/devops/vaults.gcp.env
```diff
@@ -26,7 +26,7 @@ COLIN_SVC_VERSION="op://API/$APP_ENV/colin-api-namex/COLIN_SVC_VERSION"
 SOLR_SYNONYMS_API_URL="op://API/$APP_ENV/solr-synonyms-api-gcp/SOLR_SYNONYMS_API_URL"
 SOLR_SYNONYMS_API_VERSION="op://API/$APP_ENV/solr-synonyms-api-gcp/SOLR_SYNONYMS_API_VERSION"
 LEGAL_API_URL="op://API/$APP_ENV/legal-api/LEGAL_API_URL"
-LEGAL_API_VERSION="op://API/$APP_ENV/legal-api/LEGAL_API_VERSION"
+LEGAL_API_VERSION="op://API/$APP_ENV/legal-api/LEGAL_API_VERSION_2"
```
severinbeauvais, line 29: Please update the key name here.

eve-git (approved):

ahamood1:
- @EPortman I updated the app version. Thanks for catching that. 
- @eve-git @severinbeauvais gonna merge. Thanks for the review!

EPortman:
@ahamood1 sorry I ended up cutting you in line and have bumped up the version to `1.2.43`. 
https://github.com/bcgov/namex/blob/main/api/namex/VERSION.py#L1

Do you mind quickly rebasing and updating to 1.2.44 aha?

EPortman (approved):

PR #1815, opened 2025-05-16
test namex api with keyless pubsbu (#1814)

bolyachevets:
Main goal not use API Key for pubsub authentication - since after names moved from openshift to gcp it is no longer needed.
To achieve this goal need to switch to common pubsub library (which is also a good thing), since it also has ability to pick up default credentials, such as CloudRun service account.

After namex-api change is merged, need to do the same (remove pubsub api key usage and local pubsub library in services/) in:

jobs: inprogress_update, nr-day-job, emailer
servces: namex-pay, solr-names-update, solr-feeder

Also, need to make sure that sa-api is used for services/jobs as some of them use default services accounts (these are permissions needed for service accounts: https://github.com/bcgov/bcregistry-sre/pull/252)


EPortman (approved):
Looks good. 

Also, I am not sure if this actually matters but NameX is set up to build with poetry 1.3.2.
https://github.com/bcgov/namex/blob/main/api/Dockerfile#L32

I noticed you regenerated the lock file with poetry 1.8.5. Should the Dockerfile be updated too to match that new version?

PR #1820, opened 2025-05-20
21835 add cloud tasks for 5 min grace period on approved/rejected emails

EPortman:
*Issue #, if available:* https://github.com/bcgov/entity/issues/21835

*Description of changes:*
Until now, the Emailer (a synchronous Flask service) could not “pause” between receiving a decision and sending the email. This commit:
- Introduces a [Google Cloud Tasks](https://cloud.google.com/tasks/docs) delay loop: on APPROVED / REJECTED notifications, we enqueue a task scheduled 5 minutes in the future rather than sending immediately.
  - Also cancels any existing pending task for the same NR number before scheduling the new one, ensuring only the latest decision triggers an email (if examiner rejects, but meant to approve and immediately reopens and approves only the approval email will be sent).
  - This Cloud task will push back to the emailer in the new `tasks/handle-send` endpoint (with service account based authentication)
- Preserves all existing templating, logging, and event-writing logic in the Emailer itself, while offloading the delay and retry semantics to Cloud Tasks.

This gives users a configurable grace period to change their decision before the email goes out, without blocking the Flask request cycle. 

Cloud Tasks Queue created in dev: https://console.cloud.google.com/cloudtasks/queue/northamerica-northeast1/namex-emailer-pending-send-queue-dev/tasks?inv=1&invt=Abx76A&project=a083gt-dev


EPortman:
**High Level Flow**
image

**Example of tasks in queue**
image

eve-git (approved):
LGTM

PR #1825, opened 2025-05-23
Update emailer-cd-gcp.yml

eve-git:
correct inputs reference

*Issue #, if available:*

*Description of changes:*
Got an error when running CD. Following the suggested solution:
image


eve-git:
no need to update

PR #1827, opened 2025-05-23
Ahamood1 28610 xml external expansion

ahamood1:
# Issue 28610:*

### Main change:
This change ensures that even if an attacker tries to inject malicious XML with external entity references, the parser will simply ignore them rather than attempting to resolve them, effectively preventing XXE attacks.

parser = etree.XMLParser(resolve_entities=False)
xml_content = etree.fromstring(response.content, parser)

#### Reason for Change:
- Code scan alert 232:
> Parsing untrusted XML files with a weakly configured XML parser may lead to an XML External Entity (XXE) attack. This type of attack uses external entity references to access arbitrary files on a system, carry out denial-of-service (DoS) attacks, or server-side request forgery. Even when the result of parsing is not returned to the user, DoS attacks are still possible and out-of-band data retrieval techniques may allow attackers to steal sensitive data.

### Change in Solr.py
e.message is not valid and was removed as this will just trigger another exception in the catch block. 


api/namex/resources/mras.py
```diff
@@ -36,7 +36,8 @@ def load_xml_response_content(response, xpath_query=None):
     :return:
     """
     # Parse the XML
-    xml_content = etree.fromstring(response.content)  # noqa: S320
+    parser = etree.XMLParser(resolve_entities=False)
+    xml_content = etree.fromstring(response.content, parser)  # noqa: S320
```
EPortman, line 40: You can probably remove the `noqa: S320` since you fixed this vulnerability: https://docs.astral.sh/ruff/rules/suspicious-xmle-tree-usage/

EPortman (approved):

api/namex/resources/mras.py
EPortman: Nvm lol I dont know what it wants there.
image

ahamood1:
Please hold on reviewing this. I need to dig into it a bit more now that #noqa was removed.

PR #1831, opened 2025-05-27
27245 upgrade solr admin app to flask3

EPortman:
*Issue #, if available:* https://github.com/bcgov/entity/issues/27245

*Description of changes:*
This PR upgrades the solr-admin-app from Flask 1 to Flask 3 and replaces the deprecated `flask-oidc` authentication library with `authlib`.

## Why this change was necessary:
- The `flask-oidc` package is not compatible with Flask 3.
- The `flask-jwt-oidc` library used by NameX API is designed for securing APIs, not server-rendered UIs. It doesn’t support login redirects, session-based workflows, or browser-driven authentication—features required by this Flask Admin UI.

## What was changed:
- Replaced `flask-oidc` with `authlib` to support OpenID Connect via the standard well-known configuration and JWKS endpoint.
- Implemented a custom `/oidc_callback` route to handle the auth code exchange manually and store the token in session.
- Removed the monkeypatch used to make flask-oidc work in the original implementation.

This change makes the authentication flow standards-compliant and compatible with Flask 3, while maintaining support for session-based UI access control in the admin interface.


solr-admin-app/manage.py
```diff
@@ -1,18 +0,0 @@
...
-
-app, admin = create_application()
-migrate = Migrate(app, db)
-manager = Manager(app)
-
-manager.add_command('db', MigrateCommand)
-
-if __name__ == '__main__':
-    current_app.logger.debug('Running the Manager')
-    manager.run()
```
EPortman, line 18: Removed manage.py – Flask 3 provides built-in support for database management via flask db commands

solr-admin-app/monkeypatch.py
```diff
@@ -1,80 +0,0 @@
...
-            with open(ca_certs_filename, 'wb') as file:
-                file.write(text)
-
-            # Touch our indicator so we don't add the certificate again.
-            pathlib.Path(ca_certs_already_monkeyed).touch()
-        except FileNotFoundError:
-            current_app.logger.warning(
-                'monkeypatch: httplib2 CA certificates file expected in "%s" but not found', ca_certs_filename)
-        else:
-            current_app.logger.debug('monkeypatch: httplib2 CA certificates file in "%s"', ca_certs_filename)
```
EPortman, line 80: This was a hack job to make the old auth flow work, it is no longer used in the new auth flow.

solr-admin-app/solr_admin/views/synonym_view.py
```diff
@@ -56,6 +60,7 @@ class SynonymView(SecuredView):
     def on_model_change(self, form, model, is_created):
         model.synonyms_text = _alphabetize_csv(model.synonyms_text)
         _validate_synonyms_text(model.synonyms_text)
+        model.stems_text = get_stems(model.synonyms_text)
```
EPortman, line 63: The stems_text column in the synonym table has the NOT NULL constraint, but the value for it used to be in the after_model_change hook - which runs after db insertion. This just moves it to the on_model_change to get past the NOT_NULL constraint. 

This is actually a bug in the solr-admin-app now when trying to add a new synonym (example below).
image

PR #1833, opened 2025-05-28
Ahamood1 28610 xml external expansion

ahamood1:
*Issue #28610:*

*Description of changes:*
Parsing untrusted XML files with a weakly configured XML parser may lead to an XML External Entity (XXE) attack. This type of attack uses external entity references to access arbitrary files on a system, carry out denial-of-service (DoS) attacks, or server-side request forgery. Even when the result of parsing is not returned to the user, DoS attacks are still possible and out-of-band data retrieval techniques may allow attackers to steal sensitive data.

See [Github security scan 232](https://github.com/bcgov/namex/security/code-scanning/232)


EPortman:
Whenever you update a dependency you also have to update the lock file so it is up to date / in sync (why the build check if failing and why you wouldn't be able to run this locally either).

Please run `poetry lock --no-update` on the API with `poetry 1.3.2`

Also you should be able to remove the `lxml = "^4.9.1"` dependency since it is no longer getting used (keep the `defusedxml = "^0.7.1"` dependency)

EPortman (approved):
lgtm

PR #1835, opened 2025-06-03
24162 - update Wait Time query

eve-git:
*Issue #, if available:*
https://github.com/bcgov/entity/issues/24162

*Description of changes:*
update wait time query:
        Calculate the median examination time (in days, hours, or minutes)
        between request submission and the first decision event, excluding special cases.

        The calculation logic:
        1. For each NR, select only the first decision event (action='patch' and state_cd in
           ['APPROVED', 'CONDITIONAL', 'REJECTED', 'CANCELLED']) per NR.
        2. Only include decision events that occurred yesterday or later.
        3. Exclude NRs that contain any REAPPLY action or have multiple decision making actions.
        4. Skip NRs that complete payment more than 5 days after submission (rare).
        5. Calculate and return the median waiting time using PostgreSQL percentile_cont(0.5).


EPortman (approved):
LGTM

PR #1845, opened 2025-06-13
chore: update 1pass location for search internal url change

kialj876:
*Issue #, if available:*

*Description of changes:*
- search internal url is no longer in the API vault


EPortman (approved):
lgtm, although I don't have one-pass access so cannot confirm that the password exists at that new location. 

Should the name-request vault also be updated?
https://github.com/bcgov/namerequest/blob/main/app/devops/vaults.env#L25

PR #1854, opened 2025-06-25
Add SA Identity Token to All Synonyms API Requests #29147

ahamood1:
*Issue #, if available:* [29147](https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/29147)

*Description of changes:*
- Added a new module: `api/namex/utils/synonyms_api_auth.py`
- `synonyms_api_auth.py` provides:
  - Logic to fetch the Synonyms API URL (audience)
  - A method to check whether a request's URL matches the target
  - Identity token generation using `google.oauth2.id_token`
- A custom `request` method was implemented to:
  - Match target requests
  - Inject the identity token
  - Replace the original `swagger_client.rest.RESTClientObject.request` via monkey patching


api/namex/analytics/solr.py
eve-git, line 1: why adding "return None" at line 813?

PR #1855, opened 2025-06-26
28974 - Search Pagination NR

Rajandeep98:
*Issue #, if available:* /bcgov/entity#28974

*Description of changes:*
Added hasmore flag for pagination

### DIFF
--- a/api/namex/resources/requests.py
+++ b/api/namex/resources/requests.py
@@ -960,6 +960,9 @@ class Request(Resource):
                 # add a generated comment re. this NR being reset
                 json_input['comments'].append({'comment': 'This NR was RESET.'})
 
+                # send the event to the namex emailer, to cancel the in-flight task, if there is one
+                queue_util.publish_email_notification(nrd.nrNum, 'RESET')
+
             try:
                 previousNr = json_input['previousNr']
                 if previousNr:
--- a/services/emailer/pyproject.toml
+++ b/services/emailer/pyproject.toml
@@ -45,6 +45,7 @@ extend-select = [
 
 extend-ignore = [
   "S113",  # no timeout
+  "C901"   # too complex
 ]
 
 [tool.ruff.lint.per-file-ignores]
--- a/services/emailer/src/namex_emailer/constants/notification_options.py
+++ b/services/emailer/src/namex_emailer/constants/notification_options.py
@@ -17,6 +17,7 @@ class Option(Enum):
     REJECTED = "REJECTED"
 
     CONSENT_RECEIVED = "CONSENT_RECEIVED"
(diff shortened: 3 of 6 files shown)


eve-git (approved):

stevenc987 (approved):

vysakh-menon-aot (approved):

api/namex/resources/requests.py
```diff
@@ -644,20 +644,19 @@ def post():
             and search_details.limit is not None
             and search_details.limit > 0
         ):
-            q = q.offset((search_details.page - 1) * search_details.limit).limit(search_details.limit)
-
-        q = q.offset((search_details.page - 1) * search_details.limit).limit(search_details.limit)
+            q = q.offset((search_details.page - 1) * search_details.limit).limit(search_details.limit+1)
+        q = q.offset((search_details.page - 1) * search_details.limit).limit(search_details.limit+1)
```
hfekete, line 648: This and line above that is under IF looks identical to me, if they are, do you need the IF statement and that additional statement

api/namex/resources/requests.py
```diff
@@ -644,20 +644,19 @@ def post():
...
         requests = request_auth_search_schemas.dump(q.all())
+        has_more = len(requests)> search_details.limit
         actions_array = [
             nr_filing_actions.get_actions(r['requestTypeCd'], r['entity_type_cd'], r['request_action_cd'])
-            for r in requests
+            for r in requests[:search_details.limit]
```
hfekete, line 653: This would usually be breaking change, but because the default `limit` is set to `100000` it might not affect any real use case. 
The problem might happen if someone tries to adjust the `limit`, then all the calls that do not specify limits might break. Potentially setting limit to `None` would avoid that, with comment why is None.

hfekete: Also, granted, the limit was here before too, so that means an API that was not setting limits, would have been limited to 100 001 NR without actually knowing this.

api/namex/resources/requests.py
```diff
@@ -644,20 +644,19 @@ def post():
...
         ]
         for r, additional_fields in zip(requests, actions_array):
             if additional_fields:
                 r.update(additional_fields)
-
-        return jsonify(requests)
-
+        requests = requests or []
+        return jsonify({'requests': requests[:search_details.limit], 'hasMore': has_more})
```
hfekete, line 659: Oh ! The test case is failing cause this is actually a breaking change.
Before the response was a list of NRs, 

Now it is an object `{ requests:[], has_more: bool }` this definitely should be a new version of the search endpoint as all current calls will fail.

PR #1856, opened 2025-06-30
Ahamood1 29147 add syn auth token 2

ahamood1:
*Issue [entity/29147](https://app.zenhub.com/workspaces/names-team-board-new-655554cbddd49510027dad2e/issues/gh/bcgov/entity/29147)*

*Description of changes:*

The first fix caused a TypeError got multiple values for argument query_params Root cause was mixing explicit keyword arguments like headersNone with args and kwargs in the monkey patched request method

This update fixes the issue by removing explicit keywords and instead relying solely on args and kwargs injecting the Authorization header via kwargs to avoid signature conflicts


eve-git (approved):
LGTM

PR #1857, opened 2025-07-07
29378 add status code to API exception logging in payments module

EPortman:
*Issue #, if available:* https://github.com/bcgov/entity/issues/29378

*Description of changes:*
Enhanced exception logging in the client payment module:
1. `log_api_error_response`
     - Now includes the HTTP status code (if available) and the string representation of the error.
2. `ApiClientError` and `ApiClientException`
     - `ApiClientError` now explicitly passes a status code.
     - `ApiClientException` attempts to extract .status from the wrapped error; falls back to the provided status_code if not present.
3.  'ApiAuthError'
     - Removed the unused and misleading response parameter.
     - Added a default 401 status code.
     - Simplified instantiation: the only usage now passes just the message.


EPortman:
> It might be worth logging the payload on a flag, incase we see 400's we'll have to troubleshoot

Good idea. Will ask Omid to create a FF in Launch Darkly and make the updates.

stevenc987 (approved):

PR #1968, opened 2026-05-12
32694-Fix consent received emails.

mengdong19:
*Issue #, if available:*
bcgov/entity/32694
*Description of changes:*
Set default template in case of no group returned.


stevenc987 (approved):
Looks good

PR #1970, opened 2026-05-14
fix:use correct service account env vars for solr api auth

Oluwapizzy:
Summary
- api/config.py was reading ' NAME_REQUEST_SERVICE_ACCOUNT_CLIENT_ID/SECRET' for the Solr API service account, but 'api/devops/vaults.gcp.env' injects ' SOLR_API_SERVICE_ACCOUNT_CLIENT_ID/SECRET' as the new vars

Changes
'api/config.py : updated 'SOLR_API_SERVICE_ACCOUNT_CLIENT_ID" and 'SOLR_API_SERVICE_ACCOOUNT_CLIENT_SECRET  to read from the correct env vars matching what is in 'vaults.gcp.env.


stevenc987 (commented):
Could you please increase the version number when you work for namex api?

stevenc987 (approved):

PR #1975, opened 2026-06-10
 (#32707)feat: add test-only endpoint to set NR state without Solr update

Oluwapizzy:
https://github.com/bcgov/entity/issues/32707

Adds PATCH /api/v1/namerequests/<nr_id>/test-state allowing API partners to force a name request to APPROVED or REJECTED in non-production environments without publishing a names state-change event - so the change is never propagated to Solr.

- New TestNameRequestState resource in the name_requests namespace (@jwt.requires_auth); deliberately omits send_name_request_state_msg so the update does not reach the Solr conflict index.
- APPROVED approves the first name choice and rejects alternatives; REJECTED rejects all names (mirrors a real decision).
- Guarded by ALLOW_TEST_ENDPOINT config flag (defaults false); returns 404 when not enabled - production never sets the flag.
- Wires ALLOW_TEST_ENDPOINT via vaults.gcp.env per environment.
- Adds endpoint tests: state update with no queue publish, invalid-state 400, and disabled-returns-404.


stevenc987 (approved):

stevenc987 (commented):
Looks good.

